PR #39, opened 2024-05-14
Boilerplate code for fast deploy of NDSL

FlorianDeconinck:
**Description**
In order to present an unified, easy and clean API to the user, we start here a series of "boilerplate" code that wrap common use cases. The first of those is the "quick test" with a one tile, no boundaries, no comms, for `numpy ` and `orchestrated dace:cpu`

**How Has This Been Tested?**
Adding two boilerplate code

**Checklist:**

### DIFF
--- a/ndsl/boilerplate.py
+++ b/ndsl/boilerplate.py
@@ -4,12 +4,9 @@ import numpy as np
 
 from ndsl import (
     CompilationConfig,
-    CubedSphereCommunicator,
-    CubedSpherePartitioner,
     DaceConfig,
     DaCeOrchestration,
     GridIndexing,
-    LocalComm,
     NullComm,
     QuantityFactory,
     RunMode,
--- a/ndsl/boilerplate.py
+++ b/ndsl/boilerplate.py
@@ -1,0 +1,93 @@
+import numpy as np
+from ndsl import (
+    StencilFactory,
+    DaceConfig,
+    DaCeOrchestration,
+    GridIndexing,
+    StencilConfig,
+    CompilationConfig,
+    RunMode,
+    SubtileGridSizer,
+    NullComm,
+    QuantityFactory,
+    TileCommunicator,
+    TilePartitioner,
+)
+
+from typing import Tuple
+
+
+def _get_one_tile_factory(
+    nx, ny, nz, nhalo, backend, orchestration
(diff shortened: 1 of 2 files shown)


fmalatino (approved):
lgtm

xyuan (dismissed):
would we have a util directory, and put all these utility code there instead of ndsl directory?

FlorianDeconinck:
I personally don't like `util` directories as they become a dumping ground for everything and anything quickly (we use to have one we clean up with Frank). Each tools & features should be named with what they do

xyuan:
sure, that makes sense, thanks

ndsl/boilerplate.py
```diff
@@ -0,0 +1,95 @@
...
+    RunMode,
+    StencilConfig,
+    StencilFactory,
+    SubtileGridSizer,
+    TileCommunicator,
+    TilePartitioner,
+)
+
+
+def _get_one_tile_factory(
```
oelbert, line 21: Might rename this as `get_factories_one_tile` and similarly for the others, also to expose it directly?

ndsl/boilerplate.py
```diff
@@ -0,0 +1,95 @@
...
+) -> Tuple[StencilFactory, QuantityFactory]:
+    """Build a Stencil & Quantity factory for Numpy"""
+    return _get_one_tile_factory(
+        nx=nx,
+        ny=ny,
+        nz=nz,
+        nhalo=nhalo,
+        backend="numpy",
+        orchestration=DaCeOrchestration.Python,
+    )
```
oelbert, line 95: `get_factories_one_tile_orchestrated_cpu` and `get_factories_one_tile_numpy` maybe?

oelbert (approved):
Looks good. Small suggestion if you want for names.

Is part of the thinking that we will end up making a similar function for gpu orchestrated backends and others?

ndsl/boilerplate.py
FlorianDeconinck: Made more generic and detailed why we don't expose it (simplification & the fact that the orchestration is broken between two inputs instead of one, see #46)

fmalatino (approved):

oelbert (approved):
Great, thanks!

PR #41, opened 2024-05-14
GT4Py / NDSL / Serialbox Tutorials

gmao-ckung:
Several Jupyter notebooks have been developed that step through a GT4Py and NDSL tutorial as well as Serialbox/serialization tutorials to extract data from Fortran into Python for porting and verification.


FlorianDeconinck:
This is looking good. For the fortran #2 on reading serialize data, I wonder if we should jump over the serialbox API and just demonstrated the Translate test infrastructure?

gmao-ckung:
I may be in the minority, but where can I see the Translate test infrastructure?

FlorianDeconinck:
> I may be in the minority, but where can I see the Translate test infrastructure?

Good point, we haven't pull the system down to NDSL! It's in `pyFV3`/`pySHiELD` so we will need to refactor it out

gmao-ckung:
Ok I will leave the translate tests pull to NDSL to the experts.

FlorianDeconinck:
We are rebase-ing this

PR #67, opened 2024-08-27
Multi-modal metric for 32-bit float translate tests

FlorianDeconinck:
**Description**
We propose an (optional) new metric to clear the translate test in case of 32-bit computation were the current metric seems to fail distinguishing between noise and signal.

The new metric goes as follows. Test passes if
- No NaNs AND
- Absolute difference <1e-13 (1e-10 for 32-bit float) OR
- Relative difference 0.0001% (computation is 0.0001% of the reference) OR
- ULP < 1 (the error is within the "hardware noise")

As of now this metric _does not_ include overrides. It can be triggered in the test by adding `--multimodal_metric`

And the output looks like this 

image

This PR also brings a much needed clean up of the failure report for the regular metric in translate test, removing the callstack. Looks like this:

image

**Checklist:**
- [x] I have made corresponding changes to the documentation


fmalatino (dismissed):

ndsl/testing/comparison.py
```diff
@@ -1,68 +1,265 @@
...
-        success = np.logical_or(
-            success,
-            np.logical_and(
-                np.abs(computed_data) < near_zero, np.abs(ref_data) < near_zero
-            ),
-        )
-    return success
+        return success
 
+    def __str__(self) -> str:
```
twicki, line 90: As per [docs]( https://docs.python.org/3/reference/datamodel.html#object.__repr__) :
If a class defines __repr__ but not __str__(), then __repr__() is also used when an “informal” string representation of instances of that class is required.

I think is not necessary to do this, but does not hurt

ndsl/testing/comparison.py
```diff
@@ -1,68 +1,265 @@
...
+    ):
+        super().__init__(reference_values, computed_values)
+        self.absolute_distance = np.empty_like(self.references)
+        self.absolute_distance_metric = np.empty_like(self.references, dtype=np.bool_)
+        self.relative_distance = np.empty_like(self.references)
+        self.relative_distance_metric = np.empty_like(self.references, dtype=np.bool_)
+        self.ulp_distance = np.empty_like(self.references)
+        self.ulp_distance_metric = np.empty_like(self.references, dtype=np.bool_)
+
+        self.relative_fraction = 0.000001
```
twicki, line 183: Why is this down here and not with the other class variables?

ndsl/testing/comparison.py
```diff
@@ -1,68 +1,265 @@
...
+        if self.references.dtype is (np.float32, np.int32):
+            self.absolute_eps = max(eps, self._f32_absolute_eps)
+        else:
+            self.absolute_eps = max(eps, self._f64_absolute_eps)
+        self.ulp_threshold = 1.0
```
twicki, line 188: Why is this down here and not with the other class variables?

ndsl/testing/comparison.py
```diff
@@ -1,68 +1,265 @@
...
+    ) -> npt.NDArray[np.bool_]:
+        if self.references.dtype in (np.float64, np.int64, np.float32, np.int32):
+            max_values = np.maximum(
+                np.absolute(self.computed), np.absolute(self.references)
+            )
+            # Absolute distance
+            self.absolute_distance = np.absolute(self.computed - self.references)
+            self.absolute_distance_metric = self.absolute_distance < self.absolute_eps
+            # Relative distance (in pct)
+            self.relative_distance = np.divide(self.absolute_distance, max_values)
```
twicki, line 204: we don't have the div by 0 check in here that is in the legacy. Not needed?

twicki (changes requested):

ndsl/testing/comparison.py
FlorianDeconinck: My blurry mind seem to remember an instance where this mattered. But it might be under py 2...

ndsl/testing/comparison.py
FlorianDeconinck: Well we have no denominator normalization here so no risk of div by 0. The NaNs are checked independently.

twicki (approved):
LGTM

Issue #42, opened 2024-05-14
DaCe Orchestration issues


fmalatino:
**Describe the bug**
DaCe orchestration failing when unused variables in `pySHiELD` `microphysics` module parsed. Specifically the `sedimentation` and `icloud` methods were passed unused variables, `d0_vap`, `lv00`, and `cracw` which would result in a `ValueError`. Variables have been since removed in [PR 15](https://github.com/NOAA-GFDL/PySHiELD/pull/15)

**To Reproduce**
Replace `d0_vap` and `lv00` into `sedimentation` method and calls, and `cracw` into `icloud` method and calls. Make sure to use a `dace` backend, with the `FV3_DACEMODE` environment variable set to `BuildAndRun` to enable orchestration and running of the model in `pace`. To start run use:
`mpirun -n X python -m pace.run <config yaml>`

**Expected behavior**
DaCe should be able to use orchestration regardless of variable usage.

**System Environment**
Describe the system environment, include:
- OS: RHEL 8.9
- Backend used: dace:cpu
- Environment variables set: FV3_DACEMODE=BuildAndRun
- Compiler(s): gcc/12.3.0, python/3.11, 
- MPI type, and version: openmpi/5.0.0
- netCDF Version: netcdf/4.9.2
- If this bug came from a model run, which model: baroclinic_c12_orch_cpu.yaml

FlorianDeconinck:
Another case popped up of the same issue: unused variables create a parsing issue in `orch:dace:X`.

Digging the issue lives in the gt4py/dace bridge.

In orchestration the `StencilFactory` uses a `lazy_stencil` ([code](https://github.com/NOAA-GFDL/NDSL/blob/a37f641962d3120850d61806a87b00a3703da15b/ndsl/dsl/stencil.py#L346)) to defer build at JIT time. This system refers to the `DaCeLazyStencil` ([code](https://github.com/GridTools/gt4py/blob/main/src/gt4py/cartesian/backend/dace_stencil_object.py)). 

1/ When DaCe takes over to move the code under the SDFG IR, the `__sdfg__` function gets executed. The stencil get packed into an `unexpanded` version which signature is given [here](https://github.com/GridTools/gt4py/blob/f472c9041dca338a8d73c7226ee7892ac00ab78d/src/gt4py/cartesian/backend/dace_stencil_object.py#L149) as a simple list of the declared arguments.

[...]

PR #69, opened 2024-09-04
Non-ideal fix for unreachable scalar variables

fmalatino:
**Description**
As authored by @FlorianDeconinck. Current hotfix for scalar variables which are unreachable during DaCe orchestration procedure

(HOT)Fixes # [42](https://github.com/NOAA-GFDL/NDSL/issues/42)
Hot fix until DaCe, GT4Py bridge refactor

**Checklist:**

### DIFF
--- a/ndsl/stencils/corners.py
+++ b/ndsl/stencils/corners.py
@@ -1003,6 +1003,7 @@ def fill_corners_dgrid_defn(
     with computation(PARALLEL), interval(...):
         # this line of code is used to fix the missing symbol crash due to the node visitor depth limitation
         acoef = mysign
+        x_out = x_in
         # sw corner
         with horizontal(region[i_start - 1, j_start - 1]):
             x_out = mysign * y_in[0, 1, 0]


oelbert (dismissed):
🙃

twicki (approved):
We take what we need. Can we have reverting this reflected in an issue such that we come back to it once we can?

FlorianDeconinck:
Logging the full story + examples + patch code fix that partially fixes things here: https://github.com/NOAA-GFDL/NDSL/issues/70

bensonr:
Do we need the same construct for yout/yin?

FlorianDeconinck:
> Do we need the same construct for y_out/y_in?

No, we are just forcing the stencil to not become empty when all the horizontal region are being culled. Empty stencil is not an issue but we have a signature/argument passing problem that show up when it happens.

I asked Frank to change it to `x_out=x_out` which is nonsensical but keeps the stencil from being entirely empty.

bensonr:
The acoef=mysign was put in to fix a similar if not the same issue with scalar variable removal.  Are these separate, but related issues?

FlorianDeconinck:
> The acoef=mysign was put in to fix a similar issue with scalar variable removal. Are these separate, but related issues?

Exactly... It all boils down to: we have a bridle way to link signature as described by GT4Py, real call signature in code and SDFG expectation for data going in and out. #70 has a series of 3 tests that encompasses all the issues and some code that fix some of it (but not all).

I believe the right way to deal with this is a good top to bottom clean up of the call structure, simplifying some optimization that are micro-opt with little to no performance improvement and complexify the entire system

xyuan:
this is to fix same issue but with different stencil, this fix doesn't work
when scales large scale run, it requires to rewrite the gt4py interface
optimization code to allow the void stencil, or void loop nodes for the
backend.

FlorianDeconinck:
Seems like the proper fix is not quite there. We will be merging this to unblock release.

FlorianDeconinck (approved):

PR #71, opened 2024-09-05
Support for 4d Data

oelbert:
**Description**
This PR adds support for 4D data structures in NDSL, especially the translate tests, and also fixes a bug where TranslateFortranData2Py would cast everything to Float instead of respecting types.

**How Has This Been Tested?**
The PBL and surface layer translate tests have been run with these updates

**Checklist:**


ndsl/dsl/gt4py_utils.py
```diff
@@ -154,6 +157,9 @@ def make_storage_data(
         data = _make_storage_data_2d(
             data, shape, start, dummy, axis, read_only, backend=backend
         )
+    elif n_dims == 4:
+        
+        data = _make_storage_data_4d(data, shape, start, backend=backend)
```
bensonr, line 162: Remove the empty line.

I may have missed it, but there is no test to ensure ndims <=4.  I'm guessing the gt4py.storage.from_array function will crash at that point since data is undefined.

ndsl/dsl/gt4py_utils.py
```diff
@@ -257,6 +262,25 @@ def _make_storage_data_3d(
     return buffer
 
 
+def _make_storage_data_4d(
```
FlorianDeconinck, line 265: Can we generalize this with something like
```python
def _make_storage_data_Nd(
    data: Field,
    shape: Tuple[int, ...],
    start: Tuple[int, ...],
    *,
    backend: str,
) -> Field:
    buffer = zeros(shape, backend=backend)
    idx = tuple([slice(start, start + shape_dim) for shape_dim in data.shape])
    buffer[idx] = asarray(data, type(buffer))
    return buffer
```

and we just have to pass a default start if it doesn't exist (that can be build with something like `tuple([0]*data.ndim)`

fmalatino (approved):
LGTM.

PR #74, opened 2024-09-05
Hotfix for LegacyMetric errors with 0 compute and 0 metric errors

### LINKED ISSUE #72: error comparison bug in legacymetric
**Describe the bug**
The comparison in LegacyMetric `if np.isnan(metric_err) or (metric_err > worst_metric_err):` can fail if all metric errors are negative real numbers, leading to no "worst index" being assigned and causing a memory error. It also means large negative errors won't be considered as bad as small positive ones, which is undersirable. There is a similar error if all computed vaules are 0

**To Reproduce**
Run a test where the computed outputs are all smaller than the reference outputs or 0

**Expected behavior**
metric errors should be calculated correctly and the correct worst index should be found

oelbert:
**Description**
The comparison in LegacyMetric `if np.isnan(metric_err) or (metric_err > worst_metric_err):` can fail if all metric errors are negative real numbers, leading to no "worst index" being assigned and causing a memory error. It also means large negative errors won't be considered as bad as small positive ones, which is undersirable. This PR fixes that by comparing the absolute value of the errors to find the worst.

Fixes #72 

**How Has This Been Tested?**
Physics translate tests have been run with this

**Checklist:**

### DIFF
--- a/ndsl/testing/comparison.py
+++ b/ndsl/testing/comparison.py
@@ -52,9 +52,10 @@ class LegacyMetric(BaseMetric):
         near_zero,
     ) -> npt.NDArray[np.bool_]:
         if self.references.dtype in (np.float64, np.int64, np.float32, np.int32):
-            denom = np.abs(self.references) + np.abs(self.computed)
+            denom = self.references
+            denom[self.references == 0] = self.computed[self.references == 0]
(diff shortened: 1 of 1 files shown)


FlorianDeconinck (approved):
Good catch

PR #82, opened 2024-10-24
[Serialbox to Netcdf tool] Collapse all rank into 1 if rank have different sized data

FlorianDeconinck:
**Description**
We are developing a new capacity of serialbox to append scalars to a one dimension array. This can be used to save data within iterative solver. This means that the number of data point save can vary by rank, so we have developed a solution to collapse all ranks into one to fit with the Translate test system. E.g.:
- Detect when array have different dimensions, if only one dimension, collapse
- Clean up type infer
- Log work being done

**How Has This Been Tested?**
Locally

**Checklist:**

### DIFF
--- a/ndsl/stencils/testing/serialbox_to_netcdf.py
+++ b/ndsl/stencils/testing/serialbox_to_netcdf.py
@@ -145,6 +145,7 @@ def main(
                 # feedback on one rank
                 colapse_all_ranks = False
                 data_shape = list(rank_list[0][varname][0].shape)
+                print(f"  Exporting {varname} - {data_shape}")
                 for rank in range(total_ranks):
                     this_shape = list(rank_list[rank][varname][0].shape)
                     if data_shape != this_shape:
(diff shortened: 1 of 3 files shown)


fmalatino (approved):

fmalatino (approved):

ndsl/stencils/testing/serialbox_to_netcdf.py
```diff
@@ -173,22 +185,53 @@ def main(
...
                     encoding[varname] = {"zlib": True, "complevel": 1}
+
             dataset = xr.Dataset(data_vars=data_vars)
             dataset.to_netcdf(
                 os.path.join(output_path, f"{savepoint_name}.nc"), encoding=encoding
             )
 
 
+def get_data_collapse_all_ranks(total_ranks, n_savepoints, output_list, varname):
+    if total_ranks <= 0:
```
oelbert, line 206: `<= 0`? Can we have negative total_ranks??

oelbert (approved):
👍

ndsl/stencils/testing/serialbox_to_netcdf.py
FlorianDeconinck: Probably not but I did a bot of "defensive coding" here. If we pass -1 for any reason we don't crash

PR #90, opened 2024-12-11
Fixed precision Field descriptors

FlorianDeconinck:
Describe 64/32-bit FloatFields in `dsl.typing`. Part of ground work for 32/64 bit dynamics.


fmalatino (approved):
lgtm

bensonr (approved):

PR #94, opened 2024-12-24
Global literal & temporary precision (for int & float) extended to gt4py and centralized

FlorianDeconinck:
Debugging the 32-bit GEOS model we figured out that the precision was not necessarily pass down to `gt4py` properly. We've added a `GT4PY_LITERAL

- Move precision definition to `ndsl.dsl.__init__`
- Detect bad order of imports, e.g., gt4py import before NDSL Adapt code to use the newly defined NDSL_GLOBAL_PRECISION - Set internal GT4Py literal precision definition

:warning:  This enforces order of import. Any file that starts with `import gt4py.cartesian.gtscript...` which is often the case WILL FAIL. :warning: 

The above points to a wider refactor: forward imports of `gtscript` to `ndsl` so user _don't have to import `gt4py` ever_

**How Has This Been Tested?**
Should be transparent to 64-bit cases and in GEOS test. That said, I expect many failures in `pyFV3`/`Pace` because of the order of import.

**Checklist:**

### DIFF
--- a/ndsl/__init__.py
+++ b/ndsl/__init__.py
@@ -1,3 +1,4 @@
+from . import dsl  # isort:skip
 from .comm.communicator import CubedSphereCommunicator, TileCommunicator
 from .comm.local_comm import LocalComm
 from .comm.mpi import MPIComm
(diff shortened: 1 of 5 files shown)


FlorianDeconinck:
Expecting failure here

FlorianDeconinck:
Better PR coming up

PR #104, opened 2025-01-31
Automatic Int precision and stencil regeneration change

CharlesKrop:
Added a feature to automatically determine the precision of an integer based on the global precision of the stencil. retained functionality of i32/i64, but now raises an error if an integer of the incorrect precision is passed to the stencil. In short: `Int/i32/i64` should now behave identically to `Float/f32/f64`.

Changed the default value of `rebuild` from True to False for _get_factories. This alleviates a problem of unmodified stencils being recompiled when they are called from a subclass.


ndsl/dsl/typing.py
```diff
@@ -31,26 +31,27 @@ def floating_point_precision() -> int:
...
 NDSL_32BIT_FLOAT_TYPE = np.float32
 NDSL_64BIT_FLOAT_TYPE = np.float64
+NDSL_32BIT_INT_TYPE = np.int32
+NDSL_64BIT_INT_TYPE = np.int64
 
 
-def global_set_floating_point_precision():
+def global_set_precision():
     """Set the global floating point precision for all reference
     to Float in the codebase. Defaults to 64 bit."""
```
FlorianDeconinck, line 40: Update docstring

FlorianDeconinck (changes requested):

ndsl/dsl/typing.py
```diff
@@ -22,35 +22,41 @@
...
+NDSL_32BIT_INT_TYPE = np.int32
+NDSL_64BIT_INT_TYPE = np.int64
 
 
-def global_set_floating_point_precision():
-    """Set the global floating point precision for all reference
-    to Float in the codebase. Defaults to 64 bit."""
-    global Float
-    precision_in_bit = floating_point_precision()
+def global_set_precision():
```
FlorianDeconinck, line 43:
```diff
- def global_set_precision():
+ def global_set_precision() -> type:
```

ndsl/dsl/typing.py
```diff
@@ -22,35 +22,41 @@
...
+    global Float, Int
+    precision_in_bit = get_precision()
     if precision_in_bit == 64:
-        return NDSL_64BIT_FLOAT_TYPE
+        return NDSL_64BIT_FLOAT_TYPE, NDSL_64BIT_INT_TYPE
     elif precision_in_bit == 32:
-        return NDSL_32BIT_FLOAT_TYPE
+        return NDSL_32BIT_FLOAT_TYPE, NDSL_32BIT_INT_TYPE
     else:
         NotImplementedError(
```
FlorianDeconinck, line 53:
```suggestion
        raise NotImplementedError(
```

FlorianDeconinck:
While you are in the vicinity can you also had specialization type for `IntField32/64`, `IntFieldIJ32/64`, `IntFieldK32/64`

FlorianDeconinck (approved):
LGTM

PR #117, opened 2025-02-27
Feature: NetCDF output precision configurable

fmalatino:
**Description**
This PR enables the precision of data for NetCDF output to be configurable through the `diagnostics_config` section of input config yaml as used by Pace. Precision can be set in the following way:
```
diagnostics_config:
  path: output
  output_format: netcdf
  precision: Float | float32 |float64
 ```
  The `Float` option will use whatever is the current value of the `PACE_FLOAT_PRECISION` environment variable for output precision. 
The `float32` option will set the output precision to be `numpy.float32`. 
The `float64` option will set the output precision to be `numpy.float64`. 
The default option is `Float`. 
The latter two options allow for instances when the desired run precision, as set by `PACE_FLOAT_PRECISION`, will be different than the required output precision. 
If the combination of `float64` is used with `PACE_FLOAT_PRECISION = 32`, an error will be thrown.

**How Has This Been Tested?**
Tested using the current set of unit tests in the CI

**Checklist:**

### DIFF
--- a/ndsl/monitor/netcdf_monitor.py
+++ b/ndsl/monitor/netcdf_monitor.py
@@ -1,12 +1,13 @@
 import os
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Set
+from warnings import warn
 
 import fsspec
 import numpy as np
 
 from ndsl.comm.communicator import Communicator
-from ndsl.dsl.typing import Float
+from ndsl.dsl.typing import Float, get_precision
 from ndsl.filesystem import get_fs
 from ndsl.logging import ndsl_log
 from ndsl.monitor.convert import to_numpy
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: ndsl/monitor/netcdf_monitor.py [hunk 1]
```diff
@@ -114,6 +115,7 @@ def __init__(
         path: str,
         communicator: Communicator,
         time_chunk_size: int = 1,
+        precision: str = "Float",
```
FlorianDeconinck, line 118: Why not use the type here directly rather than a string? We can take a `np` type and default to `Float` that's what we do in most subsystem

fmalatino: Ok will make the change

ndsl/monitor/netcdf_monitor.py
```diff
@@ -130,6 +132,21 @@ def __init__(
         self._time_chunk_size = time_chunk_size
         self.__writer: Optional[_ChunkedNetCDFWriter] = None
         self._expected_vars: Optional[Set[str]] = None
+        if precision == Float:
```
FlorianDeconinck, line 135: I would simplify and warn of cast here to be more flexible. Something like

```python
self._transfer_type=precision
if self._transfer_type == np.float32 and get_precision() > 32:
   warn("NetCDF save: requested 32-bit float but precision of NDSL is {get_precision()}, cast will occur with possible loss of precision")
```

FlorianDeconinck (changes requested):

ndsl/monitor/netcdf_monitor.py
FlorianDeconinck: The idea, if I follow your code, would be to just write to the given user precision, but warn when the default of NDSL is higher

fmalatino: Yes this is the intention. I will make these changes thanks!

ndsl/monitor/netcdf_monitor.py
```diff
@@ -130,6 +133,11 @@ def __init__(
         self._time_chunk_size = time_chunk_size
         self.__writer: Optional[_ChunkedNetCDFWriter] = None
         self._expected_vars: Optional[Set[str]] = None
+        self._transfer_type = precision
+        if self._transfer_type == np.float32 and get_precision() > 32:
+            warn(
+                "NetCDF save: requested 32-bit float but precision of NDSL is {get_precision()}, cast will occur with possible loss of precision"
```
FlorianDeconinck, line 139:
```suggestion
               f"NetCDF save: requested 32-bit float but precision of NDSL is {get_precision()}, cast will occur with possible loss of precision"
```

FlorianDeconinck (dismissed):
Missing f-string but we are good to go

FlorianDeconinck (approved):

Issue #32, opened 2024-03-18
Namelist parameters and many physical constants don't really belong in NDSL


oelbert:
**Is your feature request related to a problem? Please describe.**
Namelist elements and default values are specified in `ndsl/namelist.py` which means that model developers have to esid ndsl code when implementing a scheme with new runtime parameters in pyFV3 or pySHiELD. Similarly if a scheme needs a new physical constant (like a heat capacity at a new temperature) it gets implemented in `ndsl/constants.py`. Moreover it is very unclear how to set up experiments with different values, such as simulating Earth with a smaller radius.

**Describe the solution you'd like**
`namelist.py` should specify a minimal, extensible namelist class that can be inherited in Pace, pyFV3, and pySHiELD namelists.
For `constants.py` ¯\_(ツ)_/¯

romanc:
> `ndsl/namelist.py` is gone since PR [#297](https://github.com/NOAA-GFDL/NDSL/pull/297) was merged.
> 
> The constants from `ndsl/constants.py` will be moved as part of the API reorganization, i.e. issue [#243](https://github.com/NOAA-GFDL/NDSL/issues/243).

@oelbert given these steps, do you think we can close this issue or is there something you imagined that isn't covered by the new namelist handling and by moving the constants out of NDSL?

PR #122, opened 2025-03-26
[minor] GEOS version of dz_min

FlorianDeconinck:
This updates `dz_min` to be 6 for GEOS.

### DIFF
--- a/ndsl/constants.py
+++ b/ndsl/constants.py
@@ -102,6 +102,7 @@ if CONST_VERSION == ConstantVersions.GEOS:
     TFREEZE = Float(273.16)
     """Freezing temperature of fresh water [K]"""
     SAT_ADJUST_THRESHOLD = Float(1.0e-6)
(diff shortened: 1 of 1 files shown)


bensonr:
Do we want to have a dycore-only parameter in the NDSL constants file?  I assumed that was mostly reserved for physical constants.   Would it be better to have a parameters or constants file for dycore properties within pyFV3?

FlorianDeconinck:
This entire file needs a pass-over. On to of that `dz_min` should be considered a namelist parameter rather than a constants -_-

I am just moving the ball forward for the dycore stabilization on our side

bensonr:
I'm okay with allowing it to go forward as long as we have an issue to discuss and remediate later.

FlorianDeconinck:
Indeed we have it logged it here https://github.com/NOAA-GFDL/NDSL/issues/32 and I added a comment for `dz_min`

PR #143, opened 2025-05-16
Field Bundle: Functional code for 4D array

FlorianDeconinck:
**Description**
- Boilerplate code taking in a 4D Quantity
- Name/Index indexer
- Natively orchastreatable 
- Deferred type hint for data dimensions
- Unit tests

**How Has This Been Tested?**
New units tests for basic usage

**Checklist:**

### DIFF
--- a/ndsl/quantity/field_bundle.py
+++ b/ndsl/quantity/field_bundle.py
@@ -1,0 +1,141 @@
+import copy
+from typing import Any, Dict, Tuple
+from ndsl.quantity.quantity import Quantity
+from ndsl.initialization.allocator import QuantityFactory
+from ndsl.dsl.typing import Float
+
+import gt4py.cartesian.gtscript as gtscript  # isort: skip
+
+from dataclasses import dataclass
+
+# ToDo: This is 4th dimensions restricted. We need a concept
+#       of data dimensions index here to be able to extend to N dimensions
+_DataDimensionIndex = int
+_FieldBundleIndexer = Dict[str, _DataDimensionIndex]
+
+
+class FieldBundle:
+    """Field Bundle wraps a nD array (3D + n Data Dimensions) into a complex
+    indexing scheme that allows a dual name-based and index-based
+    access to the underlying memory. It is paired with the `FieldBundleType`
+    which provides a way to
+
+    WARNING: The present implementation only allows for 4D array.
+    """
+
+    _quantity: Quantity
+    _indexer: _FieldBundleIndexer = {}
+
+    def __init__(
+        self,
+        bundle_name: str,
+        quantity: Quantity,
+        mapping: _FieldBundleIndexer = {},
+        do_register_type: bool = False,
+    ):
+        if len(quantity.shape) != 4:
+            raise NotImplementedError("FieldBundle implementation restricted to 4D")
+
+        self.name = bundle_name
+        self._quantity = quantity
+        self._indexer = mapping
+        if do_register_type:
+            FieldBundleType.register(bundle_name, quantity.shape[3:])
+
+    def map(self, index: _DataDimensionIndex, name: str):
+        self._indexer[name] = index
+
+    @property
+    def quantity(self) -> Quantity:
+        return self._quantity
+
+    @property
+    def shape(self) -> Tuple[int, ...]:
+        return self._quantity.shape
+
+    def groupby(self, name: str) -> Quantity:
+        raise NotImplementedError
+
+    def __getattr__(self, name: str) -> Quantity:
+        if name not in self._indexer.keys():
+            # This replicates as close possible the default behavior of getattr
+            # without breaking orchestration
+            return None  # type: ignore
+        # ToDo: extend the dims below to work with more than 4 dims
+        assert len(self._quantity.data.shape) == 4
+        return Quantity(
+            data=self._quantity.data[:, :, :, self._indexer[name]],
+            dims=self._quantity.dims[:-1],
+            units=self._quantity.units,
+            origin=self._quantity.origin[:-1],
+            extent=self._quantity.extent[:-1],
+        )
+
+    @property
+    def __array_interface__(self):
+        return self._quantity.__array_interface__
+
+    @property
+    def __cuda_array_interface__(self):
+        return self._quantity.__array_interface__
+
+    def __descriptor__(self) -> Any:
+        return self._quantity.__descriptor__()
+
+    @staticmethod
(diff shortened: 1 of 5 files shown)


ndsl/quantity/field_bundle.py
```diff
@@ -0,0 +1,143 @@
...
+from ndsl.dsl.typing import Float
+from ndsl.initialization.allocator import QuantityFactory
+from ndsl.quantity.quantity import Quantity
+
+
+import gt4py.cartesian.gtscript as gtscript  # isort: skip
+
+
+# ToDo: This is 4th dimensions restricted. We need a concept
+#       of data dimensions index here to be able to extend to N dimensions
```
romanc, line 14: Do we already have an issue to track this work?

ndsl/quantity/field_bundle.py
```diff
@@ -0,0 +1,143 @@
...
+
+    _quantity: Quantity
+    _indexer: _FieldBundleIndexer = {}
+
+    def __init__(
+        self,
+        bundle_name: str,
+        quantity: Quantity,
+        mapping: _FieldBundleIndexer = {},
+        do_register_type: bool = False,
```
romanc, line 36: Up to preference, personally, I'd omit "do" since "register" is already a verb.

ndsl/quantity/field_bundle.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def __getattr__(self, name: str) -> Quantity:
+        if name not in self._indexer.keys():
+            # This replicates as close possible the default behavior of getattr
+            # without breaking orchestration
+            return None  # type: ignore
+        # ToDo: extend the dims below to work with more than 4 dims
+        assert len(self._quantity.data.shape) == 4
+        return Quantity(
+            data=self._quantity.data[:, :, :, self._indexer[name]],
```
romanc, line 69: up to preference
```suggestion
            data=self._quantity.data[:, :, :, self.index(name)],
```

ndsl/quantity/field_bundle.py
```diff
@@ -0,0 +1,143 @@
+import copy
+from dataclasses import dataclass
+from typing import Any, Dict, Tuple
```
romanc, line 3: since this is new code, let's use `dict` and `tuple` instead.

romanc (approved):
Proper dark magic, but that's what the DSL is here for: use whatever is needed on the inside to make it look nice on the outside.

Do I understand correctly that the current semantics are as follows:

- `FieldBundle` adds a 4th dimension,
- the 4th dimension can have any size,
- the 4th dimension is always a float (from `field_bundle.py`, line 130)

While the 4th dimension is clearly stated everywhere, the `Float` type might surprise developers.

I do have one structural question: what happens if developers initialize a `FieldBundle` without a `mapping`? Won't that break the `map()` / `index()` (and by extension `__getattr__()`) functions?

Apart from this, I think structurally, this can go in as-is for a first version. When we move to more than 4 dimensions, we can revisit the details. Might be easier once we have a bit of experience with the bundles. I left a few of nitpicks inside - nothing urgent. My wishlist (for future work) includes

- `FieldBundleType`: docstrings for the two methods and unit test (register, re-register, and type resolving with/without markup)
- `FieldBundle` docstrings

ndsl/quantity/field_bundle.py
```diff
@@ -0,0 +1,143 @@
...
+#       of data dimensions index here to be able to extend to N dimensions
+_DataDimensionIndex = int
+_FieldBundleIndexer = Dict[str, _DataDimensionIndex]
+
+
+class FieldBundle:
+    """Field Bundle wraps a nD array (3D + n Data Dimensions) into a complex
+    indexing scheme that allows a dual name-based and index-based
+    access to the underlying memory. It is paired with the `FieldBundleType`
+    which provides a way to
```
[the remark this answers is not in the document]
FlorianDeconinck: The 2 season is always less good, don't hold too high expectations

ndsl/quantity/field_bundle.py
FlorianDeconinck: We do now: https://github.com/NOAA-GFDL/NDSL/issues/145

FlorianDeconinck:
> Proper dark magic, but that's what the DSL is here for: use whatever is needed on the inside to make it look nice on the outside.
> 
> Do I understand correctly that the current semantics are as follows:
> 
> * `FieldBundle` adds a 4th dimension,
> * the 4th dimension can have any size,
> * the 4th dimension is always a float (from `field_bundle.py`, line 130)
> 
> While the 4th dimension is clearly stated everywhere, the `Float` type might surprise developers.

I fixed that. It was an actual bug because the data follows a Quantity which is not limited. Passed now as a optional field to the register. Better will read from the Quantity, but the systems of type & bundle are un-tethered for now so we have to rely on user not missing it, which was already the case before.

> I do have one structural question: what happens if developers initialize a `FieldBundle` without a `mapping`? Won't that break the `map()` / `index()` (and by extension `__getattr__()`) functions?

Then the data is "blind", which means you can still:
- Use it as a pure 4D array
- Access 3D array by index with `field.quantity[:,:,:, i]`

This is not an issue as some code don't care about the association to bespoke sub-array (for example UW moves all tracers the same). 

As for the `__getattr__` I couldn't raise like I should have because of orchestration. The next best thing is to go `None` which will fail when accessed.

> Apart from this, I think structurally, this can go in as-is for a first version. When we move to more than 4 dimensions, we can revisit the details. Might be easier once we have a bit of experience with the bundles. I left a few of nitpicks inside - nothing urgent. My wishlist (for future work) includes
> 
> * `FieldBundleType`: docstrings for the two methods and unit test (register, re-register, and type resolving with/without markup)
> * `FieldBundle` docstrings

Docstrings done.
uTest... I am going under a tunnel :)

romanc:
> > I do have one structural question: what happens if developers initialize a `FieldBundle` without a `mapping`? Won't that break the `map()` / `index()` (and by extension `__getattr__()`) functions?
> 
> Then the data is "blind", which means you can still:
> 
>     * Use it as a pure 4D array
> 
>     * Access 3D array by index with `field.quantity[:,:,:, i]`
> 
> 
> This is not an issue as some code don't care about the association to bespoke sub-array (for example UW moves all tracers the same).

Aha, I see. All good then. This makes sense to me and I think we should keep that flexibility.

romanc (approved):
LGTM!

fmalatino (approved):

PR #154, opened 2025-05-28
Optional and non-optional dependencies

romanc:
**Description**

It looks like in earlier versions, a couple packages were optional. In particular this includes the following packages

`dace`, `gt4py`, `mpi4py`, and `xarray`

which are now all installed by default given the current `setup.py`. We can thus simplify their imports (in case they were still imported conditionally). In addition, a couple code paths didn't import `cupy` from `ndsl.optional_imports`. This has been corrected too.

Doing all this, I discovered a bug in `ndsl/checkpointer/snapshots.py`, where a snapshot's dataset would only ever contain the first `savepoint`. This was unearthed by tests in `tests/checkpointer/test_snapshot.py`, which suddenly started failing after cleaning up the imports.

As a follow-up, PR https://github.com/NOAA-GFDL/NDSL/pull/156 will look into re-enabling tests with `zarr` installed. Looks like these tests will need some love/work before we can re-enable them.

**How Has This Been Tested?**

Covered by existing test cases.

**Checklist:**

### DIFF
--- a/ndsl/checkpointer/snapshots.py
+++ b/ndsl/checkpointer/snapshots.py
@@ -1,10 +1,10 @@
 import collections
 
 import numpy as np
+import xarray as xr
 
 from ndsl.checkpointer.base import Checkpointer
 from ndsl.optional_imports import cupy as cp
-from ndsl.optional_imports import xarray as xr
 
 
 def make_dims(savepoint_dim, label, data_list):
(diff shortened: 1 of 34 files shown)


FlorianDeconinck (changes requested):
Couple of extra import clean up since we are there

romanc (commented):
Some inline comments on the latest changes. New changes are only in the last commit. The rest just has new commit hashes because I rebased.

oelbert (approved):

PR #159, opened 2025-05-29
Packaging: Use of pyproject.toml and move to minimal setup.py for modernization

fmalatino:
**Description**
This PR moves the metadata and configuration for the package to a `pyproject.toml` file, as is prescribed by [Deprecate legacy setup.py develop mechanism for pip install --editable  #11457](https://github.com/pypa/pip/issues/11457). The previous `setup.py` file is now minimal. `

**How Has This Been Tested?**
Tested using the current CI workflow

**Checklist:**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,49 @@
+[build-system]
+requires = ["setuptools >= 64"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "ndsl"
+version = "2025.05.00"
+dynamic = ["dependencies"]
+requires-python = ">=3.11"
+authors = [{name = "NOAA/NASA"}]
+readme = "README.md"
+license = {file = "LICENSE.txt"}
+classifiers = [
+    "Development Status :: 2 - Pre-Alpha",
+    "Intended Audience :: Developers",
+    "License :: OSI Approved :: Apache Software License",
+    "Natural Language :: English",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.11",
+]
+
+[project.optional-dependencies]
+docs = ["mkdocs-material"]
+demos = ["ipython", "ipykernel"]
+test = ["pytest", "pytest-subtests", "coverage"]
+develop = [
+    "ndsl[docs]", 
+    "ndsl[demos]", 
+    "ndsl[test]", 
+    "pre-commit"
+]
+extras = [
+    "ndsl[docs]",
+    "ndsl[demos]",
+    "ndsl[test]",
+    "ndsl[develop]",
+]
+
+[project.scripts]
+ndsl-serialbox_to_netcdf = "ndsl.stencils.testing.serialbox_to_netcdf:entry_point"
+
+[project.urls]
+Repository = "https://github.com/NOAA-GFDL/NDSL"
+
+[tool.setuptools.packages.find]
+include = ["ndsl", "ndsl.*"]
+
+[tool.setuptools]
+include-package-data = true
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ import os
 from pathlib import Path
 from typing import List
 
-from setuptools import find_namespace_packages, setup
+from setuptools import setup
 
 
 def local_pkg(name: str, relative_path: str) -> str:
@@ -10,19 +10,6 @@ def local_pkg(name: str, relative_path: str) -> str:
     path = f"{name} @ file://{Path(os.path.abspath(__file__)).parent / relative_path}"
     return path
 
-
-docs_requirements = ["mkdocs-material"]
-demos_requirements = ["ipython", "ipykernel"]
-test_requirements = ["pytest", "pytest-subtests", "coverage"]
-
-develop_requirements = test_requirements + docs_requirements + ["pre-commit"]
-
-extras_requires = {
-    "demos": demos_requirements,
-    "develop": develop_requirements,
-    "docs": docs_requirements,
-    "test": test_requirements,
-}
 
 requirements: List[str] = [
     local_pkg("gt4py", "external/gt4py"),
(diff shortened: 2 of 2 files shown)


fmalatino:
> Two questions:
> 
> 1. Is there a way to move the non-optional dependencies also to the `pyproject.toml` and get rid of `setup.py`?
> 2. Did you test local developer workflows, e.g. installing NDSL (and gt4py/dace) editable?

@romanc 
1. Unless we want to pull `dace` and `gt4py` from their respective repositories, we will need to continue to use the dynamic dependencies definitions
2. Yes I tested this by comparison to what is done currently when running `pip install -e .[extras]` on `develop` and confirming that both generate editable installations. In both cases, to achieve and editable install of either `gt4py` and `dace` one would have to go to their respective directories and run a `pip install -e .`  with either install method.

romanc:
Thanks Frank, that sounds good.

Another question: what do you think about moving the configuration from `setup.cfg` to `pyproject.toml` too?

oelbert (approved):
:+1:

fmalatino:
> Thanks Frank, that sounds good.
> 
> Another question: what do you think about moving the configuration from `setup.cfg` to `pyproject.toml` too?

@romanc, given the current issues between `Flake8` and parsing its configuration from a `pyproject.toml` file, I vote to leave it as is, also given that we are holding on to the `setup.py` file as well. When we are able to get rid of the `setup.py` file, I think it will make sense to also get rid of the `setup.cfg` file. Thoughts?

romanc:
> > Thanks Frank, that sounds good.
> > Another question: what do you think about moving the configuration from `setup.cfg` to `pyproject.toml` too?
> 
> @romanc, given the current issues between `Flake8` and parsing its configuration from a `pyproject.toml` file, I vote to leave it as is, also given that we are holding on to the `setup.py` file as well. When we are able to get rid of the `setup.py` file, I think it will make sense to also get rid of the `setup.cfg` file. Thoughts?

I'm unaware about problems between flake8 and reading the configfile. That sounds weird to me given that pyproject.toml is the new standard. But yeah, if there are issues, we can iterate and do follow-up PRs.

Is this ready to be merged/reviewed? Asking because the PR is still marked as draft...

pyproject.toml
```diff
@@ -0,0 +1,76 @@
...
+version = "2025.05.00"
+dynamic = ["dependencies"]
+requires-python = ">=3.11"
+authors = [{name = "NOAA/NASA"}]
+readme = "README.md"
+license = "Apache-2.0"
+license-files = ["LICENSE.txt", "ndsl/viz/fv3/README.md"]
+classifiers = [
+    "Development Status :: 2 - Pre-Alpha",
+    "Intended Audience :: Developers",
```
romanc, line 16: should we add
```none
    "Intended Audience :: Science/Research",
    "Topic :: Scientific/Engineering :: Atmospheric Science",
    "Private :: Do Not Upload"
```

The last one will add another small barrier for anyone trying to upload NDSL to pypi. Not really security - just avoids someone accidentally uploading NDSL to pypi, which is - as far as I understand - what we are trying to avoid (for now at least).

pyproject.toml
```diff
@@ -0,0 +1,76 @@
+[build-system]
+requires = ["setuptools >= 77.0.3"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "ndsl"
+version = "2025.05.00"
+dynamic = ["dependencies"]
+requires-python = ">=3.11"
```
romanc, line 9: Should we add an upper bound too? For example, iirc python 3.13 won't support numpy 1.x anymore and from what I - very vaguely - remember, we don't really support 3.12 in the sense that I've heard people say things break when going there ... (which would have to be investigated and we don't really have the time for this - at least I don't)

romanc (commented):
I've been reading up on python packing and I agree, this is the most standard way add local package dependencies. It will support all current workflows and prep us for the future to come.

This is starting to look nice. To be clear, can we get rid of `setup.cfg` now? Otherwise just some minor suggestions inline.

(And sorry for the un-asked review. I was just monitoring activity and thought I'd share my thoughts / endorsement for this move to `pyproject.toml`.

pyproject.toml
FlorianDeconinck: Let's be restrictive and keep it under `<3.12` indeed.

.pre-commit-config.yaml
```diff
@@ -39,7 +39,7 @@ repos:
     - id: flake8
       name: flake8
       language_version: python3
-      args: [--config, setup.cfg]
+      args: ["--ignore=F401,E203,E501,F841,W503,E704"]
```
romanc, line 42: What do you think about using [flake8-pyproject](https://github.com/john-hen/Flake8-pyproject) to read the `flake8` configuration of the `pyproject.toml` as we do with all the other tools?

pyproject.toml
```diff
@@ -0,0 +1,80 @@
...
+
+[tool.setuptools]
+include-package-data = true
+
+[tool.flake8]
+exclude = ["docs"]
+extend-ignore = ["E203","E302","E704","F841","W293","E501","W503"]
+max-line-length = 88
+
+[tool.aliases]
```
romanc, line 59: what do you think about sorting `[tool.*]` sections alphabetically?

```toml
[tool.aliases]

[tool.coverage]
...

[tool.setuptools]

```

romanc: We could also explore usage of 

```yaml
  - repo: https://github.com/macisamuele/language-formatters-pre-commit-hooks
    rev: v2.15.0
    hooks:
      - id: pretty-format-toml
        args: [--autofix, --indent, "2"]
      - id: pretty-format-yaml
        args: [--autofix, --preserve-quotes, --indent, "2", --offset, "2"]
```

to format `yaml` (not sure if we have) and `toml` files

romanc (commented):
Coming along nicely! :rocket:

pyproject.toml
```diff
@@ -0,0 +1,80 @@
...
+    "Natural Language :: English",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.11",
+]
+
+[project.optional-dependencies]
+docs = ["mkdocs-material"]
+demos = ["ipython", "ipykernel"]
+test = ["pytest", "pytest-subtests", "coverage"]
+dev = [
```
romanc, line 28: Renaming this to `dev` (in https://github.com/NOAA-GFDL/NDSL/pull/159/commits/ee755b2947131de1b6a9bcaae8387c8947672c16) is now inconsistent with `ndsl[develop]` a few lines below. Also, the top-level `README.md` would need updating. Did you change this out of a need?

.pre-commit-config.yaml
fmalatino: I have included this as an additional dependency to the `dev` install option. Even with it added I still found I needed to add these to the pre-commit config

romanc: Ah, I think, it (also) needs to be configured in `.pre-commit-config.yaml` [as additional dependency](https://github.com/john-hen/Flake8-pyproject?tab=readme-ov-file#pre-commit-hook) of `flake8`  for usage with `pre-commit` (which is what we use on the CI). Having it as part of the `dev` extra might still make sense. in case people run `flake8` manually (or as part of their IDE) from the command line.

romanc: Looking good now :)

romanc (approved):
Nice! Thanks for bringing us up to speed with recent python config developments :rocket:

PR #165, opened 2025-06-11
Fixes for external grid data read-in method

fmalatino:
**Description**
This PR introduces fixes to the external grid data read-in method in `generation.py`. The method previously lacked a halo update and a call to `fill_corners_2d`.

**How Has This Been Tested?**
These changes have been tested outside the CI via the test located in Pace at `pace/tests/mpi/ext_grid`, as well as with a comparison of the output from a c12 baroclinic run using internally generated grid data and the external data.

**Checklist:**

### DIFF
--- a/ndsl/grid/generation.py
+++ b/ndsl/grid/generation.py
@@ -482,6 +482,12 @@ class MetricTerms:
         terms._grid_64.view[:, :, 0] = rad_conv * x
         terms._grid_64.view[:, :, 1] = rad_conv * y
 
+        terms._comm.halo_update(terms._grid_64, n_points=terms._halo)
+
+        fill_corners_2d(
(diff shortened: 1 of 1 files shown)


fmalatino:
> So it looks good to me for a code viewpoint. I still don't really understand why we have to transpose things.
> 
> @oelbert / @fmalatino : at the very least we should have a docstring in this file or in a readme that spills out your findings

@FlorianDeconinck I do not think we *need* to tranpose things, this is done to match the layout already used by Pace. I can amend this if we are indifferent regarding what the layout should be.

fmalatino:
> I can't judge of the scientific validity of this, since I have no idea what comes in, but code wise is ok

@FlorianDeconinck these changes only bring in functionality for reading in longitude and latitude values generated by external sources; scientific validity is assumed to be validated within the generator.

bensonr (approved):

PR #171, opened 2025-06-20
Store static eta files under `tests/data`

fmalatino:
**Description**

This PR removes the need to generate eta files for running tests. These files are actually static (i.e. they never change) and can thus be stored as netCDF under `tests/data/`. This simplifies running tests both locally (for users) and on the CI.

**How Has This Been Tested?**
The current workflow has been amended to reflect these changes

**Checklist:**


romanc (changes requested):
I was thinking we could use `fixtures` from `pytest` for this. I was thinking we could extend the [tmp_path](https://docs.pytest.org/en/6.2.x/tmpdir.html) fixture, create only the needed data on the fly and use [fixture finalization](https://docs.pytest.org/en/6.2.x/fixture.html#teardown-cleanup-aka-fixture-finalization) to clean up the files.

What is suggested in this PR only works if tests run in order. We should never rely on test order, especially since we run tests in parallel. Also note that the currently existing test already cleans up the eta file (which only works okay because there's one (and not multiple) tests in that file.

fmalatino:
> I was thinking we could use `fixtures` from `pytest` for this. I was thinking we could extend the [tmp_path](https://docs.pytest.org/en/6.2.x/tmpdir.html) fixture, create only the needed data on the fly and use [fixture finalization](https://docs.pytest.org/en/6.2.x/fixture.html#teardown-cleanup-aka-fixture-finalization) to clean up the files.
> 
> What is suggested in this PR only works if tests run in order. We should never rely on test order, especially since we run tests in parallel. Also note that the currently existing test already cleans up the eta file (which only works okay because there's one (and not multiple) tests in that file.

I have amended my approach and re-pushed

tests/grid/test_eta.py
```diff
@@ -56,7 +32,8 @@ def test_set_hybrid_pressure_coefficients_correct(km):
     directly from the NetCDF file.
     """
 
-    eta_file = Path.cwd() / f"eta{km}.nc"
+    eta_file = str(Path.cwd()) + "/ndsl/data/" + f"eta{km}.nc"
+    eta_data = xr.open_dataset(eta_file)
```
romanc, line 36: `xarray` has native support for `pathlib` objects. No need to stringify (and hardcode the separator to "/"), i.e. you can do things like

```py
path = Path.cwd() / "tests" / "data" / "eta" / f"{resolution}km.nc"
eta_data = xr.open_dataset(path)
```

romanc (changes requested):
Some questions

Can we put the new eta files under `tests/data` or does that interfere with MSD workflows? We could do a directory structure like

```none
tests/data/eta/README.md
tests/data/eta/79km.nc
tests/data/eta/79km-non_mono.nc
tests/data/eta/91km.nc
```

That way, all test data files are grouped together and the mono and non-mono versions of the 79km resolution are close together. Bonus points for adding a README file explaining what eta files are :wink:

`ndsl/data/__pycache__/eta.cpython-311.pyc` shouldn't be in there, right?

romanc (dismissed):
Nice! Thanks everybody for their patience while we did iterations on this PR.

romanc (approved):
Ship it! :rocket:

oelbert (approved):
LGTM

PR #175, opened 2025-06-27
tests: re-enable caching tests and more cleanups in tests/dsl

romanc:
**Description**

This PR looks at tests under `tests/dsl` and does some cleanup work. In particular

- Re-enable tests in `tests_caches.py`: They were skipped because `MPI` is always defined (since it's a non-optional import)
- Fix tests in `tests_caches.py`: Tests haven't been running since we moved away from python 3.8 (because the python version was hard-coded in there ...)
- Move `make_storage` into a utility module (used from two test modules)
- Run tests in `tests/dsl/test_stencil_factory.py` with more than one backend
- Adding type hints and "inlining" single parameter sets.

Just to disambiguate: Some tests (the ones inside `tests/mpi`) rely on `MPI` being changed to `None` in case of a serial run:

https://github.com/NOAA-GFDL/NDSL/blob/2986b450386b5006d847f246ff6e8b23abdc9190/tests/mpi/mpi_comm.py#L1-L6

Tests outside `test/mpi` import `MPI` from `mpi4py` directly and thus check the world size themselves. Might be worth cleaning this up anyway ... or - you know - use `pytest.marker` instead. Feel free to leave this open if you want to discuss this.

**How Has This Been Tested?**

Running tests locally.

**Checklist:**

### DIFF
--- a/ndsl/stencils/testing/conftest.py
+++ b/ndsl/stencils/testing/conftest.py
@@ -376,7 +376,7 @@ def parallel_savepoint_cases(
 
 def pytest_generate_tests(metafunc):
     backend = metafunc.config.getoption("backend")
-    if MPI is not None and MPI.COMM_WORLD.Get_size() > 1:
+    if MPI.COMM_WORLD.Get_size() > 1:
         if metafunc.function.__name__ == "test_parallel_savepoint":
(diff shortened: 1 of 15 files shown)


FlorianDeconinck (approved):
LGTM

PR #177, opened 2025-07-02
[Tool] Best Guess Netcdfs diff

FlorianDeconinck:
## Main change

This PR proposes a new tool to check NetCDFs against one another with a series of guess to match compute domains. The tool describe it's logic as it goes down and fail if it can't resolve.

Options for name mapping & halo size.

After (re)installing NDSL, it can be called with

```bash
best_guess_diff /path/to/A.nc4 /path/to/B.nc4 --halo 3 --name_mapping mapper.yaml
```

where `name_mapping` is a yaml file with a single dictionary mapping name in A to B.

## Minor QOL

- Debugger now knows how to FieldBundle
- `quantity.to_netcdf` when no rank is given looks at the rank itself
- `ndsl_log` is imported early to avoid circular dependencies

### DIFF
--- a/ndsl/stencils/testing/best_guess_diff.py
+++ b/ndsl/stencils/testing/best_guess_diff.py
@@ -1,0 +1,194 @@
+import argparse
+
+import xarray as xr
+import yaml
+import numpy as np
+import pathlib
+
+
+def get_parser():
+    parser = argparse.ArgumentParser(
+        "Attempt to diff two NetCDFs with similar data."
+        "Differences that can be reconcialed are strict domain vs halo, variable name mapping."
+        "They program will report on assumptions taken."
+    )
+    parser.add_argument(
+        "netcdf_A",
+        type=str,
+        help="path of NetCDFs, named A in the logs.",
+    )
+    parser.add_argument(
+        "netcdf_B",
+        type=str,
+        help="path of NetCDFs, named B in the logs.",
+    )
+    parser.add_argument(
+        "-nm",
+        "--name_mapping",
+        type=str,
+        help="[Optional] Yaml file describing the mapping of the names.",
+    )
+    parser.add_argument(
+        "-ha",
+        "--halo",
+        type=int,
+        default=3,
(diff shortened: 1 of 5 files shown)


ndsl/stencils/testing/best_guess_diff.py
```diff
@@ -0,0 +1,194 @@
...
+
+import numpy as np
+import xarray as xr
+import yaml
+
+
+def get_parser():
+    parser = argparse.ArgumentParser(
+        "Attempt to diff two NetCDFs with similar data."
+        "Differences that can be reconcialed are strict domain vs halo, variable name mapping."
```
oelbert, line 12:
```suggestion
        "Differences that can be reconciled are strict domain vs halo, variable name mapping."
```

ndsl/stencils/testing/best_guess_diff.py
```diff
@@ -0,0 +1,194 @@
...
+        "They program will report on assumptions taken."
+    )
+    parser.add_argument(
+        "netcdf_A",
+        type=str,
+        help="path of NetCDFs, named A in the logs.",
```
oelbert, line 18:
```suggestion
        help="path of first NetCDF file, named A in the logs.",
```

ndsl/stencils/testing/best_guess_diff.py
```diff
@@ -0,0 +1,194 @@
...
+            ulp_diff,
+            dims=[f"GP_{ulp_diff.shape[0]}"],
+        )
+
+        print("  [Success]")
+
+    xr.Dataset(dataset).to_netcdf(f"best_guest_diff_{pathlib.Path(netcdf_A).stem}.nc4")
+
+
+def entry_point():
```
oelbert, line 182: too good to call it `main()`? :joy:

oelbert (changes requested):
Small changes, but also how many of these did you run into when diffing NetCDFs? Do you want to add checks for extra dimensions while we're at it

fmalatino (approved):

PR #184, opened 2025-07-21
Upgrade `mpi4py` to >4 

FlorianDeconinck:
The current work to run at scale on Grace Hopper unified silicon ARM64 hardware as showcased the need to move `mpi4py` above 4x.

The original restriction to `3.1.5` was done out of an abundance of caution in a moment where halo exchange  were questioned.

Alas, the new allocator that ships with cuda 12.5 for GH boxes especially trips the old `mpi4py`. We need to move up.

The risk is limited as `4.1` showcase a very large amount of fixes to main version 4.

### DIFF
--- a/.github/workflows/unit_tests.yaml
+++ b/.github/workflows/unit_tests.yaml
@@ -10,14 +10,20 @@ concurrency:
 jobs:
   ndsl_unit_tests:
     runs-on: ubuntu-latest
-    container:
-      image: ghcr.io/noaa-gfdl/miniforge:mpich
     steps:
 
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
           submodules: 'recursive'
+
+      - name: Setup Python 3.11
+        uses: actions/setup-python@v5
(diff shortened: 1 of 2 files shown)


fmalatino (approved):

romanc (commented):
Sounds reasonable. NDSL now logs a warning

```none
   <frozen importlib._bootstrap>:241: RuntimeWarning: suspicious MPI execution environment
  Your environment has OMPI_COMM_WORLD_SIZE=6 set, but mpi4py was built with MPICH.
  You may be using `mpiexec` or `mpirun` from a different MPI implementation.
```

and pace tests complain about the wrong number of ranks. Something seems off ... Might be in the CI configuration, but suspicious nevertheless

Also, pyFV3 is not running "Orchestrated dace-cpu Accoustics" anymore (and thus really, really fast)

FlorianDeconinck:
Go for it

fmalatino (approved):

PR #199, opened 2025-08-12
tests: fix NaNs introduced into reference values

romanc:
**Description**

PR #74 introduced better comparison in the `LegacyMetric`. Unfortunately, this overrides the reference values under certain conditions. The regression was realized because NaNs were copied into the reference values where 0 was expected.

This PR fixes the `LegacyMetric` calculation to copy the reference data before manipulating the array to avoid devision by 0 errors. In case the reference value is 0 (and thus the denominater would be 0), we set the denominator to 1 because we expect the computed value to be (close to) 0. In that case (abs(computed - expected) / 1.0) is a good value for the error being reported.

**How Has This Been Tested?**

Running tests locally.

**Checklist:**


romanc:
@oelbert any concerns merging this?

Issue #64, opened 2024-08-09
NDSL should provide a framweork for physics namelists instead of specifying the namlist members and default values for specific physics models


oelbert:
**Is your feature request related to a problem? Please describe.**

Currently namelist.py provides the complete structure for a model namelist, including the physics settings. This means that 
1. as we develop the Physics we need to also create a branch of NDSL to use in parallel,
2. the namelist is tailored to specific models instead of providing a general platform, and
3. the namelist in NDSL is getting pretty bloated as more physics is integrated.
 
This is not a sustainable way to develop a model and breaks the concept of NDSL as a middleware platform.

**Describe the solution you'd like**

NDSL should provide the infrastructure to create namelist configs for model components like a NamelistFactory and then provide that for the dycore. Then the physics schemes like pySHiELD can import NamelistFactory to create PhysicsConfig. Pace itself can then import these and compose them into a single expected namelist.

**Describe alternatives you've considered**

Alternative ideas are welcome

jjuyeonkim:
I have a draft design document. Please let me know what you think.

[20250723_DD_NDSLNamelistRefactor.pdf](https://github.com/user-attachments/files/21394825/20250723_DD_NDSLNamelistRefactor.pdf)

fmalatino:
Linking [Issue 32](https://github.com/NOAA-GFDL/NDSL/issues/32)

romanc:
Linking issue #6

romanc:
I had a read of the design document and the initial implementation above. I like how this is coming together. A few remarks/questions

- Consider renaming `BaseConfig` -> `Config`.
- Consider what should go in the base class for configuration objects. For example, I understand all subclasses should implement `validate()` because it is called in the post initialization hook. For `from_f90nml()` and `from_yaml()` are you planing on providing default implementations? If not, I wouldn't force sub-classes of `BaseConfig` to implement them. I think sub-classes should be able to choose which `from_*` functions they want to provide. If you can lift the functionality of `from_yaml()` and `from_f90nml()`, consider moving them out of the class into helper functions.

Assuming the above two, we could consider moving config-related things into a package. I could see a directory structure like this

```none
ndsl/config/__init__.py
ndsl/config/config.py      // <- contains the base class
ndsl/config/utils.py       // <- contains utilities like `from_yaml()`
```

Code style: prefer `dict[str, int]` over `Dict[str, int]`. In recent python versions, `dict` is generic and `from typing import Dict` can be omitted.

[...]

PR #200, opened 2025-08-12
Namelist Refactor: NDSL Config

jjuyeonkim:
**Description**
This is the first of multiple PRs to refactor the NDSL Namelist so that submodules have more ownership over configuration parameters and corresponding default values. This base Config class is meant to provide more structure and consistency to submodule configuration classes. PyFV3's DynamicalCoreConfig and PyShield's PhysicsConfig are meant to inherit from it. In later PRs, NDSL's Namelist and NamelistDefaults will later be pared down and possibly removed.

Fixes: https://github.com/NOAA-GFDL/NDSL/issues/64 (partial refactor)

**How Has This Been Tested?**
Basic unit tests have been added to PyFV3 tests/main. Pace unit tests also pass as expected.

**Checklist:**


ndsl/config.py
```diff
@@ -0,0 +1,121 @@
...
+    Base class for configuration objects.
+    Provides subclasses with base methods for loading from YAML and
+    Fortran namelists files.
+    """
+
+    dt_atmos: int = DEFAULT_INT
+    days: int = DEFAULT_INT
+    hours: int = DEFAULT_INT
+    minutes: int = DEFAULT_INT
+    seconds: int = DEFAULT_INT
```
jjuyeonkim, line 28: These are likely not needed here. They were initially put in because they're used by the DynamicalCoreConfig to calculate the n_steps, but it turns out that n_steps may not be used anyway by the DynamicalCoreConfig.

ndsl/config.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        if self.npx != self.npy:
+            raise ValueError(f"npx({self.npx}) and npy({self.npy}) should be equal")
+
+        positive_vars = [
+            "dt_atmos",
+            "days",
+            "hours",
+            "minutes",
+            "seconds",
```
jjuyeonkim, line 113: Should remove these too

ndsl/config.py
```diff
@@ -0,0 +1,121 @@
...
+        """Uses a Namelist to create a Config"""
+        namelist_dict = namelist_to_flatish_dict(f90_namelist.items())
+        namelist_dict = {
+            key: value
+            for key, value in namelist_dict.items()
+            if key in cls.__dataclass_fields__  # type: ignore
+        }
+        config = cls(**namelist_dict)
+        config.validate()
+        return config
```
jjuyeonkim, line 47: 1. Instead of flattening, we want to keep the structure of the namelist file.
2. We want each submodule's configuration to know which sections to pull the data from.

ndsl/config.py
```diff
@@ -0,0 +1,121 @@
...
+                        flat_config[subkey] = subvalue
+            else:
+                if key == "nx_tile":
+                    flat_config["npx"] = value + 1
+                    flat_config["npy"] = value + 1
+                elif key == "nz":
+                    flat_config["npz"] = value
+                else:
+                    if key in config.__dataclass_fields__.keys():
+                        flat_config[key] = value
```
jjuyeonkim, line 78: Similar to the from_f90nml, we want to preserve the structure of the yaml (don't flatten it) and use only relevant sections to initialize.

ndsl/config.py
jjuyeonkim, line 28: These probably shouldn't be here. They're used in the dycore config to calculate n_steps, but that should actually be removed too.

PR #205, opened 2025-08-19
NDSL Documentation via MkDocs

gmao-ckung:
**Description**
This branch contains files to generate documentation based on the docstrings within NDSL files through the use of MkDocs.


romanc (approved):
Thanks for setting this up, Chris. I think this is good as a first version. Basically, this just adds everything we have. We might want to reconsider what to show. I'd be nice, if we could steer users towards functions they should use (e.g. public API) and away from implementation details deep down in modules.

I've added two commit as review:

1. optional dependencies moved to ` pyproject.toml`  with PR https://github.com/NOAA-GFDL/NDSL/pull/159
2. I locally checkout the branch and when running the local server, I fixed a couple warnings that were easy to fix. The console is now flooded with missing argument types (which is a good thing, we want to get them out of the way eventually anyway), so it's hard to miss things.

As said above, we should probably have a conversion if we want to have modules starting with `_` (usually means they are "private" and not part of the public API) in the docs at all. Let's merge this first and have the discussion afterwards. We anyway don't (auto-) publish docs yet, so imo it's safe to merge increments.

PR #208, opened 2025-08-25
quick fix for 2D fields

oelbert:
**Description**
Sometimes we serialize data with shape (x, y, 1) and want to compare to a 2D ndsl field, this removes the extra dimension on the comparison field by default.

**How Has This Been Tested?**
Translate tests run with this.

**Checklist:**

### DIFF
--- a/ndsl/testing/comparison.py
+++ b/ndsl/testing/comparison.py
@@ -28,6 +28,9 @@ class BaseMetric:
     ):
         self.references = np.atleast_1d(reference_values)
         self.computed = np.atleast_1d(computed_values)
+        if len(self.computed.shape) == 2 and len(self.references.shape) == 3:
+            if self.references.shape[-1] == 1:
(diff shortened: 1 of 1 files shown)


fmalatino (approved):

FlorianDeconinck (changes requested):
This should be done on the `netcdf` generation size rather than a post-op on the NDSL.

Fair for now - but please comment in code why we are doing this, so it's seen as the workaround it is.

romanc:
+1 on what Florian said. Please add an issue wherever 'on the netcdf generation side' is to build this properly. Thenlink the issue here sucht that we have a trace. 

Messing with the reference data this way is imo prone to bize us back down the road. Does this happen rare enough that we could log something at INFO level that we messed with the reference data? Not that anyone would care about that lev2l of logging. Just one more place to say that this is supposed to be different.

PR #211, opened 2025-09-03
[Orchestration] Common cast operation replacements

FlorianDeconinck:
Replace `Float` and `Int` with a DaCe cast node in orchestration to go around the lack of `TypeAlias` inference in the DaCe parser.

This allows for easier orchestration going forward by automatically parsing the `Float` / `Int` keywords both in stencils and in glue code.


ndsl/dsl/dace/replacements.py
```diff
@@ -0,0 +1,32 @@
+"""This module uses DaCe's op_repository feature to override symbols/AST object
+during parsing and replace them with an SDFG compatible representation. This
+allow custom NDSL system to be natively orchestratable."""
```
romanc, line 3: Nitpick
```suggestion
"""This module uses DaCe's op_repository feature to override symbols/AST objects
during parsing and replace them with an SDFG compatible representation. This
allows custom NDSL {syntax? or symbols?} to be natively orchestratable."""
```

romanc (approved):
Looks good to me. Any way we could programmatically test this?

FlorianDeconinck:
> Looks good to me. Any way we could programmatically test this?

Yes and no. I can build a SDFG from a small bit of code that has `Float` and go check that the `cast` op has been dropped proper. But the fix is for cases where that doesn't done auto, which it is sometime! So not easy...

FlorianDeconinck:
Will move this code in `M2`

PR #212, opened 2025-09-04
[Translate] Allow for generic slicing of Nth dims

FlorianDeconinck:
And undo the bespoke `names_4D` from AI2 days


romanc:
Is it okay to leave this here for a week or so until we untangle the M2 branch?

If we need it as a breaking change for Geos v11.4, we might want to have a separate branch to accumulate related changes into and then merge with the Geos v11. 4 update.

FlorianDeconinck:
I think I can fabricate a backward comptabile system or at least fix the model side of things. Let's leave it here for now - you know you can pull the changes locally if you need them

FlorianDeconinck:
Super-seeded by #273

PR #220, opened 2025-09-10
Remove `FrozenCompiledSDFG` to avert the hidden behavior on orchestration call

twicki:
The `FrozenCompiledSDFG` was created to cache in the C pointers of the data with an eye of getting the most speed for online code into larger Fortran models. This creates a hidden (nearly undocumented) behavior in which swapping memory (by reallocation or else) is prohibited and will result in wrong answers.

We revert to the previous behavior which is safe (but slow) and test properly for future optimization.

Side quest: we fix the (rare) case where orchestration doesn't touch any `stencils`, where the `dace` build folder set in the frozen stencil call is not enough. Documented the lackluster system in code.

### DIFF
--- a/ndsl/dsl/dace/build.py
+++ b/ndsl/dsl/dace/build.py
@@ -128,6 +128,7 @@ def set_distributed_caches(config: DaceConfig):
 
     # Set read/write caches to the target rank
     from gt4py.cartesian import config as gt_config
+    import dace
 
     if config.do_compile:
         verb = "reading/writing"
@@ -135,6 +136,14 @@ def set_distributed_caches(config: DaceConfig):
         verb = "reading"
 
     gt_config.cache_settings["dir_name"] = get_cache_directory(config.code_path)
+
+    dace.Config.set(
+        "default_build_folder",
+        value="{gt_root}/{gt_cache}/dacecache".format(
+            gt_root=gt_config.cache_settings["root_path"],
+            gt_cache=gt_config.cache_settings["dir_name"],
+        ),
+    )
     ndsl_log.info(
         f"[{orchestration_mode}] Rank {config.my_rank} "
         f"{verb} cache {gt_config.cache_settings['dir_name']}"
--- a/ndsl/dsl/dace/build.py
+++ b/ndsl/dsl/dace/build.py
@@ -1,6 +1,8 @@
 from typing import List, Optional, Tuple
 
+from dace import config as dace_conf
 from dace.sdfg import SDFG
+from gt4py.cartesian import config as gt_config
 
 from ndsl.dsl.caches.cache_location import get_cache_directory, get_cache_fullpath
 from ndsl.dsl.dace.dace_config import DaceConfig, DaCeOrchestration
@@ -127,9 +129,6 @@ def set_distributed_caches(config: DaceConfig):
             )
 
     # Set read/write caches to the target rank
-    import dace
-    from gt4py.cartesian import config as gt_config
-
     if config.do_compile:
         verb = "reading/writing"
     else:
@@ -137,13 +136,18 @@ def set_distributed_caches(config: DaceConfig):
 
     gt_config.cache_settings["dir_name"] = get_cache_directory(config.code_path)
 
-    dace.Config.set(
+    # NOTE: In the (rare) case we orchestrate code _without_ any stencils, we need
+    # to set the build folder. The other code is in FrozenStencil and deals with the
+    # case of `dace` used in both orchestrated and not orchestrated.
+    # A better build system would deal with this in BOTH cases.
+    dace_conf.Config.set(
         "default_build_folder",
         value="{gt_root}/{gt_cache}/dacecache".format(
             gt_root=gt_config.cache_settings["root_path"],
             gt_cache=gt_config.cache_settings["dir_name"],
         ),
     )
+
     ndsl_log.info(
         f"[{orchestration_mode}] Rank {config.my_rank} "
         f"{verb} cache {gt_config.cache_settings['dir_name']}"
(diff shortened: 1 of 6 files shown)


ndsl/dsl/dace/build.py
```diff
@@ -127,6 +127,7 @@ def set_distributed_caches(config: DaceConfig):
             )
 
     # Set read/write caches to the target rank
+    import dace
```
romanc, line 130: Why are we having import statements in the middle of the code?

tests/dsl/orchestration/test_call.py
```diff
@@ -0,0 +1,50 @@
...
+        orchestrate(obj=self, config=stencil_factory.config.dace_config)
+        self.stencil = stencil_factory.from_dims_halo(_stencil, [X_DIM, Y_DIM, Z_DIM])
+
+    def __call__(self, out_qty):
+        self.stencil(out_qty)
+
+
+def test_memory_reallocation(factories):
+    qty_factory = factories[1]
+    code = OrchestratedProgram(factories[0], qty_factory)
```
romanc, line 38: Is this really better than the imo much simpler

```suggestion
def test_memory_reallocation():
    stencil_factory, quantity_factory = get_factories_single_tile_orchestrated(
        5, 5, 2, 0
    )
    code = OrchestratedProgram(stencil_factory, quantity_factory)
```

ndsl/dsl/dace/orchestration.py
```diff
@@ -217,12 +217,12 @@ def _build_sdfg(
 
         with DaCeProgress(config, "Loading"):
             sdfg_path = get_sdfg_path(dace_program.name, config, override_run_only=True)
+            if sdfg_path is None:
+                raise ValueError("Couldn't load SDFG post build")
```
romanc, line 221: Wouldn't it be simpler to change `get_sdfg_path()` to either return a path or raise? From what I can see, the only time we actually return `None` over there is when we are cautiously guarding against bad usage.

ndsl/dsl/dace/orchestration.py
```diff
@@ -237,7 +237,17 @@ def _call_sdfg(dace_program: DaceProgram, sdfg: SDFG, config: DaceConfig, args,
         with DaCeProgress(config, "Run"):
             if config.is_gpu_backend():
                 _upload_to_device(list(args) + list(kwargs.values()))
-            res = config.loaded_precompiled_SDFG[dace_program]()
+
+            # NOTE: this will go over declared arguments and closure arguments.
+            # It is a very slow piece of code. Compiled SDFG comes with a "fast_call"
+            # function that expects all pointers to have been made C worthy. This is
+            # something we did with "FrozenSDFG" but we undid because it meant that
```
romanc, line 244: That one is important because `FrozenSDFG` is a different concept in the gt4py/dace bridge. Over there it means that origin/domain were "frozen" into the SDFG (which is otherwise unfrozen).

```suggestion
            # something we did with "FrozenCompiledSDFG" but we undid because it meant that
```

twicki (commented):
Looks good, thanks a lot!

ndsl/dsl/dace/build.py
```diff
@@ -135,6 +136,14 @@ def set_distributed_caches(config: DaceConfig):
...
 
     gt_config.cache_settings["dir_name"] = get_cache_directory(config.code_path)
+
+    dace.Config.set(
+        "default_build_folder",
+        value="{gt_root}/{gt_cache}/dacecache".format(
+            gt_root=gt_config.cache_settings["root_path"],
+            gt_cache=gt_config.cache_settings["dir_name"],
+        ),
+    )
```
romanc, line 146: How is this related? Do we have to move a cache dir to make this work?

romanc (changes requested):
Sorry - many questions and suggestions... I think we should delete `FrozenCompiledSDFG` from `dace_config.py` and that reference

https://github.com/NOAA-GFDL/NDSL/blob/51874161dbf32b24a1c54550019b637920145c2d/ndsl/dsl/dace/orchestration.py#L200

needs to go too (otherwise we will be confused in the future).

ndsl/dsl/dace/build.py
FlorianDeconinck: Fixed

ndsl/dsl/dace/orchestration.py
FlorianDeconinck: We rely on that None:

```python
    sdfg_path = get_sdfg_path(dace_program.name, config)
    if sdfg_path is None:
        is_compiling = (
            True if DEACTIVATE_DISTRIBUTED_DACE_COMPILE else config.do_compile
        )

        if not is_compiling:
            # We can not parse the SDFG since we will load the proper
            # compiled SDFG from the compiling rank
            return None

        with DaCeProgress(config, f"Parse code of {dace_program.name} to SDFG"):
            sdfg = dace_program.to_sdfg(
                *args,
                **dace_program.__sdfg_closure__(),
                **kwargs,
                save=False,
                simplify=False,
                validate=False,  # TODO: should we have a "debug flag" to turn this on?
            )
```

ndsl/dsl/dace/build.py
FlorianDeconinck: Not related to the PR - but fix another nasty rare case, documented in code by
```python
# In the (rare) case we orchestrate code _without_ any stencils, we need
# to set the build folder    
```

Description updated.

romanc (approved):
Looking good now from my side. The cache directory / relocation tests are fragile. I had issues with them in https://github.com/NOAA-GFDL/NDSL/pull/221 too. The build config is a global that we touch in many places. I'm not sure how well this works with the existing fixtures in the relocations tests (which are my fault :see_no_evil:). If the relocation tests work on their own but don't pass when you run all tests (make sure to clean the cache), then the relocation tests are borked and could be skipped to be stabilized later.

PR #228, opened 2025-09-16
BREAKING CHANGE: change constructor of `QuantityFactory`

romanc:
# Description

This PR changes the constructor of `QuantityFactory` to initialize from a `GridSizer` (as previously) and a backend name (new). This hides the previous `numpy` argument, which is effectively an internal allocator that users shouldn't need to know about. The new constructor is what `from_backend()` is providing now. 

Usage of the old constructor was deprecated with https://github.com/NOAA-GFDL/NDSL/pull/231, which was shipped with the `2025.10.00` release. We are now changing the constructor and deprecating `from_backend()` (which is scheduled for removal in the next release because it's now redundant with the new constructor).

This is a BREAKING CHANGE and users will need to update their codes if they used the old constructor of `QuantityFactories` and didn't rely on `QuantityFactory.from_backend()`. For users relying on the `boilerplate` module, no changes need to happen. Here's a rundown of what needs to change in user-code:

In case you used `QuantityFactory.from_backend()` before, this is still valid code. With a deprecation note, we suggest to move to the new constructor, which is very similar:

```patch
- QuantityFactory.from_backend(sizer, backend)
+ QuantityFactory(sizer, backend=backend)
```

In case you were using `numpy` as allocator before, use the `"numpy"` backend

```patch
- QuantityFactory(sizer, numpy)
+ QuantityFactory(sizer, backend="numpy") 
```

In case of allocation through `cupy`, we suggest to move to the `dace:gpu` backend

```patch
- QuantityFactory(sizer, cupy)
+ QuantityFactory(sizer, backend="dace:gpu")
```

## How has this been tested?

Existing tests have been updated to use the new syntax.

## Checklist

### DIFF
--- a/ndsl/initialization/allocator.py
+++ b/ndsl/initialization/allocator.py
@@ -47,25 +47,6 @@ class QuantityFactory:
 
         self._allocator = _Allocator(self.backend)
 
-    @classmethod
-    def from_backend(cls, sizer: GridSizer, backend: str) -> QuantityFactory:
-        """Initialize a QuantityFactory to use a specific GT4Py backend.
-
-        Note: This method is deprecated. Please change your code to use the
-        constructor instead.
-
-        Args:
-            sizer: GridSizer object that determines the array sizes.
-            backend: GT4Py backend name used for performance-optimized allocation.
-        """
-        warnings.warn(
-            "QuantityFactory.from_backend(sizer, backend) is deprecated. Use "
-            "QuantityFactory(sizer, backend=backend) instead.",
-            DeprecationWarning,
-            stacklevel=2,
-        )
-        return cls(sizer, backend=backend)
-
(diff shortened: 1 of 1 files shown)


romanc:
I think we should go the way of giving our users a heads-up because imo `QuantityFactory` is a central piece of code. People should know and not suddenly run into a runtime error after the update. PR https://github.com/NOAA-GFDL/NDSL/pull/231 will add the deprecation warning of the current constructor of the `QuantityFactory`. It has tests to make sure we only trigger the deprecation warning if the constructor is called by users directly and not if the constructor is called by `QuantityFactory.from_backend()`.

We expect the next release to come in a time frame such that this PR can just sit and wait for it. As soon as we branch off for the next release, we can merge here (after another round of rebase because PR https://github.com/NOAA-GFDL/NDSL/pull/231 is likely to bring conflicts and tests will need to be adapted).

PR #239, opened 2025-09-23
Add domain-size checks on FrozenStencil

twicki:
# Description

Since our Quantity-Factory generally allocates more memory than strictly needed (for gt4py-compatibility as well as memory alignment) as can be seen [here](https://github.com/NOAA-GFDL/NDSL/blob/680d014e3ae4075ba44c3933c27e7d3f8b5c393c/ndsl/initialization/sizer.py#L155-L162), if a `Quantity`'s expected domain-size is too small for the execution domain of a stencil, it can be that we are not warned properly. 

This PR adds safe-guards around these calls to make sure we are not reading from memory that is technically allocated but never properly used. 

## How has this been tested?

New test has been added

## Checklist

### DIFF
--- a/ndsl/dsl/stencil.py
+++ b/ndsl/dsl/stencil.py
@@ -534,8 +534,6 @@ class FrozenStencil(SDFGConvertible):
         """
         all_args_as_kwargs = dict(zip(self._argument_names, tuple(list(args)))) | kwargs
 
-        domain_sizes = dict(zip((), self.domain))
-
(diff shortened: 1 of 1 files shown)


FlorianDeconinck (approved):
LGTM

Let's make sure we capture a thorough ticket for the rest of the work and link it here. Top of my head:
- clean up FV3 and/or option to run _strict_ and fail instead of warn
- extract extent analysis from gt4py

PR #240, opened 2025-09-24
Implementation of plain-numpy corners

twicki:
# Description

After a deep-dive into how our model performs on CPU it became apparent that the CopyCorners codes were underperforming by a large margin. This PR provides a plain numpy-based implementation that is numerically identical and allows for more speed-up.

The fact that the performance is low  is mostly due to the fact that corner fixes are not actually a stencil motif and are translated into very inefficient code through GT4Py into SFGD-land. Bypassing this inefficient expression of the computational motif allows for more streamlined code to be generated.

With this implementation we are deprecating the old way of doing corners - which in an update to PyFV3 should no longer be called anywhere and should be removed in the next release. 

## How has this been tested?

Ran orchestrated & non-orchestrated regression tests (CopyCorners / FvTp2d and Del6VtFlux)

## Checklist

### DIFF
--- a/ndsl/__init__.py
+++ b/ndsl/__init__.py
@@ -6,7 +6,7 @@ from .comm.null_comm import NullComm
 from .comm.partitioner import CubedSpherePartitioner, TilePartitioner
 from .constants import ConstantVersions
 from .dsl.caches.codepath import FV3CodePath
-from .dsl.dace.dace_config import DaceConfig, DaCeOrchestration, FrozenCompiledSDFG
+from .dsl.dace.dace_config import DaceConfig, DaCeOrchestration
 from .dsl.dace.orchestration import orchestrate, orchestrate_function
 from .dsl.dace.utils import (
     ArrayReport,
--- a/ndsl/boilerplate.py
+++ b/ndsl/boilerplate.py
@@ -1,7 +1,3 @@
-from typing import Tuple
-
-import numpy as np
-
 from ndsl import (
     CompilationConfig,
     DaceConfig,
@@ -16,18 +12,17 @@ from ndsl import (
     TileCommunicator,
     TilePartitioner,
 )
-from ndsl.optional_imports import cupy as cp
 
 
 def _get_factories(
     nx: int,
     ny: int,
     nz: int,
-    nhalo,
+    nhalo: int,
     backend: str,
     orchestration: DaCeOrchestration,
     topology: str,
-) -> Tuple[StencilFactory, QuantityFactory]:
+) -> tuple[StencilFactory, QuantityFactory]:
     """Build a Stencil & Quantity factory for a combination of options.
 
     Dev Note: We don't expose this function because we want the boilerplate to remain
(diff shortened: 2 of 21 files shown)


FlorianDeconinck (approved):
We need to mention - if not `assert` or `try/catch` in the model - that those are halo=3 tuned. Previous code was as well and that wasn't really said

FlorianDeconinck:
Let's move this inside `pyFV3` and deal with it in https://github.com/NOAA-GFDL/PyFV3/pull/87

ndsl/dsl/stencil_config.py
```diff
@@ -168,7 +168,7 @@ def from_dict(cls, data: dict):
 class StencilConfig(Hashable):
     compare_to_numpy: bool = False
     compilation_config: CompilationConfig = CompilationConfig()
-    dace_config: Optional[DaceConfig] = None
+    dace_config: DaceConfig = None
```
romanc, line 171: should this be
```suggestion
    dace_config: DaceConfig | None = None
```

twicki: it is never `None` as it is always set in the `__post_init__`.

romanc: ah, I see. In that case how about just `dace_config: DaceConfig` (without the `= None`)?

twicki: If you know of a way to have an optional `kwarg` that is set in the `__post_init` if not passed in that would be nice. But not setting it does not work

romanc: ah okay ... then how about we leave it as is? If it needs a default value and that default value is `None` then technically `DaceConfig | None` is the right type, no? It's also unrelated to the deprecation of copy corners ...

romanc: I mean, looking at the code, we actively rely on that `None` in `__post_init__` to distinguish the case when it has been set and when it hasn't.

romanc: Then let's make it a proper default instead of patching things up in `__post_init__`, e.g.

```patch
 @dataclasses.dataclass
 class StencilConfig(Hashable):
     compare_to_numpy: bool = False
-    compilation_config: CompilationConfig = CompilationConfig()
-    dace_config: Optional[DaceConfig] = None
     verbose: bool = False
+    compilation_config: CompilationConfig = CompilationConfig()
+
+    # We need a DaceConfig to know if orchestration is part of the build system
+    # but we can't hash it very well (for now). The workaround is to make
+    # sure we have a default Python orchestrated config.
+    dace_config: DaceConfig = DaceConfig(
+        communicator=None,
+        backend=CompilationConfig().backend,
+        orchestration=DaCeOrchestration.Python,
+    )
 
     def __post_init__(self):
         self.backend_opts = {
@@ -178,16 +186,6 @@ class StencilConfig(Hashable):
         }
         self._hash = self._compute_hash()
 
-        # We need a DaceConfig to know if orchestration is part of the build system
-        # but we can't hash it very well (for now). The workaround is to make
-        # sure we have a default Python orchestrated config.
-        if self.dace_config is None:
-            self.dace_config = DaceConfig(
-                communicator=None,
-                backend=self.compilation_config.backend,
-                orchestration=DaCeOrchestration.Python,
-            )
-
     @property
     def backend(self):
         return self.compilation_config.backend
```

(and maybe move this to a separate PR because it's unrelated to deprecating corners)

PR #241, opened 2025-09-24
[Fix] DaCe build system should be tied more closely to existing configuration

FlorianDeconinck:
# Description

GT4Py as a flag `GT4PY_COMPILE_OPT_LEVEL` that directs `-O` compiler flags. We have tied it in to the compilation level for DaCe within their CMake system. But we haven't tied it fully and this PR fixes that.

In a subsequent PR we should introduce a NDSL level configuration for optimization that drives both the GT4Py and DaCe sub-configuration.


romanc (approved):
Looking forward to building this properly in the future.

PR #250, opened 2025-10-06
State features: `fill` and data dimensions support.

FlorianDeconinck:
_Continuing the feature set for `State` as we met them._

- Introduce a `fill` method that will fill _all_ fields with a particular value
- Allow for field to be used easily with data dimensions via a temporary `Sizer` swap

Side task:
- Clean up a `global_config` that isn't used anymore.

### DIFF
--- a/ndsl/quantity/state.py
+++ b/ndsl/quantity/state.py
@@ -92,6 +92,7 @@ class State:
     def empty(
         cls,
         quantity_factory: QuantityFactory,
+        *,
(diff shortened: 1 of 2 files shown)


CharlesKrop (approved):

romanc:
Ah and I forgot to mention: would a `full` (as [in numpy](https://numpy.org/doc/stable/reference/generated/numpy.full.html)) be useful to allocation and initialize to any given number (not just 0 and 1)?

PR #252, opened 2025-10-07
gt4py update: absolute K indexing (experimental)

romanc:
# Description

This PR updates the `GT4Py` dependency of NDSL. The new version brings absolute indexing (in the K-dimension) as an experimental feature. Usage of this feature looks like

```py
@stencil(backend="dace:cpu", externals={K4: "4"})
def absolute_k_indexing(in_field: Field[np.float64], index_field: Field[IJ, np.int64], out_field: Field[np.float64], index: int) -> None:
        from __externals__ import K4

        # access third level in K (useful for fixed boundary layers)
        out_field = in_field.at(K=2)

        # works with externals
        out_field = in_field.at(K=K4)

        # also works if level is not known at compile time
        out_field = in_field.at(K=index)

        # and even if we read from another field
        out_field = in_field.at(K=index_field)
```

You can also access data dimensions (should your field have them) as `field.at(K=5, ddim=[2, 3])`.

:warning: Absolute K indexing is shipped as experimental feature. Experimental features come with the following disclaimer. Absolute K indexing

- might break science code in unforeseen ways,
- might not be supported in all backends (in fact it's only supported in `debug` and `dace:*` backends),
- might change at any point in time without prior warning, even in breaking ways,
- and might be replaced with a more suitable (set of) feature(s) or be entirely removed without replacements.

## How has this been tested?

Absolute K indexing is a gt4py feature and comes with tests at the gt4py level.

## Checklist

- [x] New check tests, if applicable, are included (downstream)


twicki (approved):
Great, thanks for updating

PR #256, opened 2025-10-07
Add `QuantityFactory.full`

### LINKED ISSUE #166: Add a "full" (or similar) method for creating Quantities with an arbitrary fill value
**Is your feature request related to a problem? Please describe.**
Sometimes a `Quantity` needs to be initialized at creation time with a value other than `0` or `1`.  Right now, `Quantity`s can be created using `.empty`, `.zeros` or `.ones`.  If a value other than zero or one is desired, an extra line is needed after creation to set the `Quantity` to the desired value.

**Describe the solution you'd like**
Numpy has a `.full` method for creating and initializing arrays with an arbitrary fill value.  For example, you can create an array of integers initialized to -1:
``` python
ktop = np.full(shape=(im, jm), fill_value=-1, dtype=int)
```

To maintain an intuitive API for users, a `.full` method could be added for creating `Quantity`s.  For example:
``` python
ktop: Quantity = quantity_factory.full(
    dims=[X_DIM, Y_DIM],
    fill_value=-1,
    units="index",
    dtype=int,
)
```
**Describe alternatives you've considered**
The alternative is to use the existing behavior by calling the `.zeros` method and then making an assignment:
``` python
ktop: Quantity = quantity_factory.empty(
    dims=[X_DIM, Y_DIM],
    units="index",
    dtype=int,
)
ktop.field[:, :] = -1
```

This works, but is not as clean and tidy.

**Additional context**
This feature would be applicable in many use cases and enjoy widespread use.

FlorianDeconinck:
Mimics `numpy.full`. Closes #166 

Also:
- some docstrings

### DIFF
--- a/ndsl/initialization/allocator.py
+++ b/ndsl/initialization/allocator.py
@@ -78,6 +78,9 @@ class QuantityFactory:
         dtype: type = Float,
         allow_mismatch_float_precision: bool = False,
     ) -> Quantity:
+        """Allocate a Quantity - values are random.
+
+        Equivalent to `numpy.empty`"""
         return self._allocate(
             self._numpy.empty, dims, units, dtype, allow_mismatch_float_precision
         )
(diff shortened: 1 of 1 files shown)


romanc (approved):
Nice 🚀

christopherwharrop-noaa (approved):
This is great!  Thank you.

PR #266, opened 2025-10-15
Locals

FlorianDeconinck:
### Locals

_AS: All of the above is geared to, and only to, `orch:dace:X` - but the system will be deployed at NDSL level in a backend-blind way._

A limitation to our optimization pipeline is the inability to differentiate "global" fields from "transient". Take the following code:

```python
class PhysicsParam:

    def __init__(self):
        ...
        self._local_field = quantity_factory.zeros()

    def __call__(self, input_A, output_B):

        self.stencil(input_A, self._local_field)
        ...
        self.stencil(self._local_field, output_B)

class Driver:
    def __init__(self):
        ...
        self.physics = PhysicsParam()

    def __call__(self, input_A, output_B):
        physics(input_A, output_B)


if __name__ == "__main__"
    d = Driver()
    d()
```

In this configuration the field `self._local_field` is technically "temporary" to `PhysicsParam` code and should therefore be `transient` in DaCe SDFG lingo. Multiple problems arise:

- how to allow the user to flag the `Quantity` as `transient`. Should we? This is both an highly technical feature, but it has an explainable usage.
- there is a difference of behavior between backends. `orch:X` backends will have memory irrelevant in `self._local_field` since DaCe would take over it. All other backends would still have the last value used

To address those, we introduce a concept of `Local` which is a field that is **forbidden** to be used outside of the module it was defined in. In the previous example `self._local_field` would become a `Local`. This `Local` would pass the `transient` flag to the `dace.Data` descriptor and open the way for futher optimization.

From this, we need to implement a few systems:

- A way to forbid access of `Locals` outside of the module that allocated it. E.g.:

```python
if __name__ == "__main__":
    d = Driver()
    d()
    print(d.physics._local_field) # <-- ILLEGAL, this is LOCAL to PhysicsParam
```

- A way to detect `Quantity` allocation within "NDSL" code to warn about potential under-optimized use.

Below is a proposal - which has been tested for orchestration:

```python
from ndsl import orchestrate, DaceConfig
import inspect
import warnings
import os
import numpy as np

# ===============  NDSL OLD CONCEPTS (summarized) =========== #


class Quantity:
    def __init__(self) -> None:
        self.data = np.empty((2, 2))
        self.transient = False

    def __str__(self) -> str:
        return "I am quantity"


# ===============  NDSL NEW CONCEPTS =========== #


class Local(Quantity):
    """This could NOT be used: with the __post_init__
    we could flip switches ourselves on quantities and make it all
    automagical without the need for a Local class"""

    def __init__(self) -> None:
        super().__init__()
        self.transient = True

    def __str__(self) -> str:
        return "I am a Local"


TOP_LEVEL: object | None = None


class NDSLRuntime:
    def __init__(self, dace_config: DaceConfig) -> None:
        self._dace_config = dace_config

    def __init_subclass__(cls, **kwargs):
        """WARNING: no code outside the `init_decorator` this is cls
        function, it will be called ONLY ONCE for monkey-patching the
        Class - not the instance !"""

        def init_decorator(previous_init):
            def new_init(self, *args, **kwargs):
                global TOP_LEVEL
                if TOP_LEVEL is None:
                    TOP_LEVEL = self
                previous_init(self, *args, **kwargs)
                self.__post_init__()

            return new_init

        cls.__init__ = init_decorator(cls.__init__)

    def __post_init__(self):
        # Check quantity allocation of NDSLRuntime supervised code
        if TOP_LEVEL == self:

            def check_for_quantity(object_: object):
                for key, value in object_.__dict__.items():
                    if isinstance(value, Quantity) and not isinstance(value, Local):
                        warnings.warn(
                            f"{type(self).__name__}.{key} is a Quantity instead of a Locals"
                            " on a NDSLRuntime - our eyebrows are frowned."
                        )
                    elif isinstance(value, NDSLRuntime):
                        check_for_quantity(value)

            check_for_quantity(self)

        orchestrate(
            obj=self,
            config=self._dace_config,
        )

    def __getattribute__(self, name):
        attr = super().__getattribute__(name)
        # We look at the direct caller frame for our own `self`
        # in the locals.
        # Any other case are forbidden.
        if isinstance(attr, Local):
            caller_frame = inspect.currentframe().f_back
            if (
                not caller_frame
                or "self" not in caller_frame.f_locals
                or not isinstance(caller_frame.f_locals["self"], type(self))
            ):
                raise RuntimeError("Locals called outside!")

        return attr


# ===============  USER CODE =========== #


class NestedCode(NDSLRuntime):
    def __init__(self, dace_config: DaceConfig) -> None:
        super().__init__(dace_config)
        self.tmp = Local()
        self.tmp_as_qty__baaaad = Quantity()

    def __call__(self):
        self.tmp.data[:] = 12


class Code(NDSLRuntime):
    def __init__(self, dace_config: DaceConfig) -> None:
        super().__init__(dace_config)
        self.nested_code = NestedCode(dace_config)

    def __call__(self):
        self.nested_code()


os.environ["FV3_DACEMODE"] = "BuildAndRun"
c = Code(DaceConfig(None, "dace:cpu_kfirst"))
c()


print("Illegal use of a Local: ", c.nested_code.tmp)  # ILLEGAL
c.nested_code.tmp.data[:] = 6  # ILLEGAL
```

Remained to be solved:
- [ ] API for Local allocation (`LocalFactory`? Method on `NDSLRuntime` taking a `QuantityFactory` ? ... ?)

PS: This also introduces a `NDSLRuntime` base class, which can be used for a better `orchestration`, `debug` and overall gives consistency and structure to NDSL code

### DIFF
--- a/ndsl/dsl/ndsl_runtime.py
+++ b/ndsl/dsl/ndsl_runtime.py
@@ -2,8 +2,8 @@ from __future__ import annotations
 
 import inspect
 import warnings
-from typing import Any
 from collections.abc import Callable
+from typing import Any
 
 from ndsl.dsl.dace import DaceConfig, orchestrate
 from ndsl.dsl.typing import Float


ndsl/quantity/temporaries.py
```diff
@@ -0,0 +1,24 @@
+import dataclasses
+
+from ndsl.quantity import Quantity, State
+
+
+@dataclasses.dataclass
+class Temporaries(State):
+    """Base class to collect temporaries Quantities.
```
romanc:
```suggestion
    """Base class to collect temporary Quantities.
```

romanc (commented):
Have you considered making a temporary quantity, i.e.

```py
class Temporary(Quantity):
    def __init__(...):
        super().__init__(...)
        # Note: I wouldn't expose `_transient` in the constructor
        # of `Quantity`  because it will only raise questions.
        self._transient = True
```

That would still allow to group all temporaries in a `State`  object, e.g.

```py
@dataclass
class MyTemporaries(State):
    tmp_a: Temporary = dataclass.field( ... )
    tmp_b: Temporary = dataclass.field( ... )
```

It would open the possibility of `State` objects with `Quantity`  and `Temporary` members, which we can't do with the current proposal. That might or might not be a good idea. I'm not familiar enough with what atmospheric scientist expect from the "State" concept. My CS-background finds it weird that I have to group my temps and non-temps separate states.

FlorianDeconinck:
PR updated with code sample & explanation of the issue/remedy

ndsl/dsl/ndsl_runtime.py
```diff
@@ -0,0 +1,126 @@
+from __future__ import annotations
+
+import inspect
+import warnings
+from typing import Any, Callable
```
romanc: `typing.Callable` is a deprecated alias for `collections.abc.Callable` since python 3.9. Let's try to avoid it. In PR #270 I've updated all usage. I haven't found a way to enforce it by CI yet, so we'll have to catch it in code reviews until then.

```suggestion
from collections.abc import Callable
from typing import Any
```

ndsl/dsl/ndsl_runtime.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+_TOP_LEVEL: object | None = None
+
+
+class NDSLRuntime:
+    """Base class to tool runtime code, allows use of Locals, orchestration and
+    debug tools.
+
+    The __call__ function will automatically orchestrated."""
```
romanc:
```suggestion
    The __call__ function will automatically be orchestrated."""
```

ndsl/dsl/ndsl_runtime.py
```diff
@@ -0,0 +1,126 @@
...
+                obj=self,
+                config=self._dace_config,
+            )
+            print(type(self))
+
+    def __getattribute__(self, name: str) -> Any:
+        attr = super().__getattribute__(name)
+        # We look at the direct caller frame for our own `self`
+        # in the locals.
+        # Any other case are forbidden.
```
romanc:
```suggestion
        # Any other case is forbidden.
```

ndsl/initialization/allocator.py
```diff
@@ -202,7 +224,7 @@ def _allocate(
             )
         except TypeError:
             data = allocator(shape, dtype=dtype)
-        return Quantity(
+        quantity = Quantity(
```
romanc: undo now that the local flag is captured in `Local` vs. `Quantity`?

ndsl/dsl/ndsl_runtime.py
FlorianDeconinck: Deep inside of me, I knew it... I was lazy to look around for the correct thing.

ndsl/quantity/state.py
```diff
@@ -60,7 +60,11 @@ def _init_recursive(cls: Any) -> dict:
 
                     initial_quantities[_field.name] = quantity_factory_allocator(
                         _field.metadata["dims"],
-                        _field.metadata["units"],
+                        (
+                            _field.metadata["units"]
+                            if "units" in _field.metadata.keys()
+                            else "unspecified"
+                        ),
```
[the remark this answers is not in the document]
FlorianDeconinck: Orthogonal to this PR but I think `units` shouldn't be required.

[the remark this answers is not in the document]
FlorianDeconinck: Reverted

fmalatino (approved):

PR #272, opened 2025-10-22
Refactor: QuantityFactory data dimensions operations

FlorianDeconinck:
# Description

The `set_extra_dim_lengths` function of `QuantityFactory` is badly named: it updates the dictionary of data dimensions, rather than set. 

Usage of this function is also a bit odd, you mostly want to _add_ dimensions. But because a `QuantityFactory` can be shared (should be shared) across many codes you want to make sure to not override previous dimensions set.

Introducing:
- `QuantityFactory.add_data_dimensions`: check for existing names and error out.
- `QuantityFactory.update_data_dimensions`: update existing or add new ones.

Deprecating `set_extra_dim_lengths` for later removal.

+ Unit tests and proper type hints

## How has this been tested?

New utest.

## Checklist


ndsl/initialization/allocator.py
```diff
@@ -52,8 +53,45 @@ def set_extra_dim_lengths(self, **kwargs: Any) -> None:
         """
         Set the length of extra (non-x/y/z) dimensions.
         """
+        ndsl_log_on_rank_0.warning(
+            "`QuantityFactory.set_extra_dim_lengths` is deprecated. "
+            "Use `add_data_dimensions` or `update_data_dimensions`.",
+        )
```
romanc: Is it possible to derive this warning from the python standard `DeprecationWarning`? Doing so would allow us to leverage the ecosystem and - for example - write tests like

```py
with pytest.deprecated_call():
    # deprecated code here
    pass
```

which would make sure that the deprecation warning actually happens.

FlorianDeconinck: So yes, but then we escape our logger. And with our logger I couldn't get it to work with it. So I felt conflicted. I can move it to `DeprecationWarning`

ndsl/initialization/allocator.py
```diff
@@ -52,8 +53,45 @@ def set_extra_dim_lengths(self, **kwargs: Any) -> None:
...
+        data_dimension_descriptions: dict[str, int],
+    ) -> None:
+        """
+        Update the length of extra (non-x/y/z) dimensions, unknown data dimensions
+        will be added, existing updated.
+
+        Args:
+            data_dimension_descriptions: Dict of name/length pairs
+        """
+        self.sizer.extra_dim_lengths.update(data_dimension_descriptions)
```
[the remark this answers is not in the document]
FlorianDeconinck: Good idea. But this severally complicates things due to the `SubtileGridSizer.from_tile_params` function using `extra_dim_lengths` as a parameter name.

We are in luck as it seems it's _always_ used outside of NDSL as a `kwarg`. So I have update this pretty deep API to force `data_dimensions`/`extra_dim_lengths` to be a `kwarg`. Let the utest god be kind to us next.

PR #276, opened 2025-10-23
Deprecate usage of ndsl.Namelist

romanc:
# Description

This PR deprecates usage of `ndsl.Namelist`. I also tried to deprecate `ndsl.NamelistDefaults` but since we don't use it as a class (i.e. it's not a dataclass and we just have it as class to group things), there's no static constructor to (easily) intercept.

Context for the deprecation is issue #64 and @jjuyeonkim continued work in pyFV3, pySHiELD, and pace to get rid of the class. If we deprecate now (before the "M2 release"), we can remove the class as soon as it is not needed anymore in any of the downstream repositories.

## How has this been tested?

New unit test checking for the deprecation message.

## Checklist

  N/A
  N/A
  No, but that's the point here.
  N/A

### DIFF
--- a/ndsl/namelist.py
+++ b/ndsl/namelist.py
@@ -1,4 +1,5 @@
 import dataclasses
+import warnings
 from typing import Any, Self
 
 import f90nml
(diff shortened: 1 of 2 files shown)


jjuyeonkim (approved):
Looks good to me.

PR #290, opened 2025-10-28
Type tweak: Allow ZarrMonitor to take in Comm | DummyComm | None

jjuyeonkim:
# Description

Small proposed type adjustment for the Zarr Monitor. Without this, there are type linting errors when attempting to update the submodules for Pace: https://github.com/NOAA-GFDL/pace/pull/156

## How has this been tested?
CI passes, linting passes in Pace locally with this update

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (e.g. add new modules to docs/docstrings/)
- [ ] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] New check tests, if applicable, are included

PR #258, opened 2025-10-09
ci: all the types (or at least most of them) enforced with `mypy`

romanc:
# Description

This PR is stacked PR on top of https://github.com/NOAA-GFDL/NDSL/pull/257 with even more type hints. pySHiELD tests are currently failing because of what this PR surfaced, see

- https://github.com/NOAA-GFDL/PySHiELD/pull/65
- https://github.com/NOAA-GFDL/NDSL/pull/259

Once this is merged, we can probably close issue https://github.com/NOAA-GFDL/NDSL/issues/47. Further clean-up will be needed, but this should be a good start.

## How has this been tested?

All good as long as CI is green.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
  N/A
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published in downstream modules
- [x] New check tests, if applicable, are included

PR #257, opened 2025-10-08
ci: stricter type checking with `mypy`

romanc:
# Description

This PR updates `mypy` (16 -> 18) and changes the value of  `strict_optional` from `false` to its default value `true`. Doing so will propagate `None` types and enforce proper checks are done before attributes/methods are accessed on an object that is potentially of `None` type.

According to `mypy`'s docs, having `strict_optional = "false"` in your config file [is evil](https://mypy.readthedocs.io/en/stable/config_file.html#none-and-optional-handling) because it hides a bunch of problems that the type checker could  flag.

This PR depends on

- https://github.com/NOAA-GFDL/NDSL/pull/260 which fixes the remaining issues in `ndsl/global_config.py` by deleting the file
- https://github.com/NOAA-GFDL/NDSL/pull/240 which fixes the remaining issue in `ndsl/dsl/stencil.py`

The proper cleanup types follows in https://github.com/NOAA-GFDL/NDSL/pull/258, which is stacked on top of this PR.

## How has this been tested?

`mypy` is now strict about objects that are potentially `None`.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation: N/A
- [x] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published in downstream modules: N/A
- [x] New check tests, if applicable, are included

Issue #157, opened 2025-05-28
Snapshots and the zarr dependency


romanc:
# Description

Snapshots were supposed to an iteration over the translate test system, allowing in-code "serialbox behavior". The (optional) `zarr` dependency is loosely tied to that concept as a means to add some compression capability.

The snapshot system (and it's `zarr` monitor) are both unused at the moment. Tests relying on the optional `zarr` are currently skipped due to the uninstalled dependency. Naively [re-enabling these tests](https://github.com/NOAA-GFDL/NDSL/pull/156) (by just installing the optional dependency), shows that most of these tests fail. 

This will be useful in the future and as such this issue serves as a placeholder for a time when we have a real use-case and/or the capacity to work on it.

PR #292, opened 2025-10-29
fix: allow any Comm object in ZarrMonitor

romanc:
# Description

This PR is fallout from adding types in PR #257 and #258. The `ZarrMonitor` provides a `DummyComm` which is instantiated in case no `Comm` object is given. The type of the `Comm` object in `ZarrMonitor` was wrongly limited to that `DummyComm`, which only broke when we attempted to update the submodule in `pace`.

This PR is what @jjuyeonkim proposed in https://github.com/NOAA-GFDL/NDSL/pull/290#discussion_r2471167863. 

## How has this been tested?

Tested locally by updating NDSL in pace and running `pre-commit`. 

## Checklist

  No new checks added because the `ZarrMonitor` is a bit in limbo, see https://github.com/NOAA-GFDL/NDSL/issues/157. There aren't any `ZarrMonitor` tests now and adding one would mean to add `zarr` as a test dependency, which I think isn't justified just to add two simple tests that would check usage with `DummyComm` and with a regular `Comm` object.

### DIFF
--- a/ndsl/comm/__init__.py
+++ b/ndsl/comm/__init__.py
@@ -5,7 +5,7 @@ from .caching_comm import (
     CachingRequestReader,
     CachingRequestWriter,
 )
-from .comm_abc import Comm, Request
+from .comm_abc import Comm, ReductionOperator, Request
 
 
 __all__ = [
@@ -15,5 +15,6 @@ __all__ = [
     "CachingRequestReader",
     "CachingRequestWriter",
     "Comm",
+    "ReductionOperator",
     "Request",
 ]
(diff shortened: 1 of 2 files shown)


twicki (approved):
It makes sense to finish the API of the comm. I am not 100% sure if the implementation of all of these can raise and not just do nothing. But if this works, I'm happy!

romanc:
> in the case of a dummy-communicator where all the data is already centralized on one rank, doing nothing is (presumably) the right thing and now we error out if it is used instead. Obviously the "right" thing to do is to properly re-implement all the methods in memory with copies, but that is a lot of work and as far as I know the extent of this calculator, it is used for gathering data and for exchaning halo, both of which are no-ops in single rank mode

Follow-up issue: https://github.com/NOAA-GFDL/NDSL/issues/294

PR #296, opened 2025-10-30
[feature] Common data types for orchestration via `compiletime`

FlorianDeconinck:
# Description

DaCe likes static memory arrays as parameters. Anything more involved will fail to parse. There are two ways to go around this:

- Use `op.repo` or `__descriptor__` to list out to DaCe what it needs. We are exploring those, but it's not straighforward
- Use `dace.compiletime` and move description read to the first access

That latter solution, though imperfect, has been used until now via an API that requires understanding of orchestration. Here we propose to move it internally for the common types: `Quantity` & `State`.

This also helps `NDSLRuntime` work more seamlessly since it orchestrate `__call__` automatically.

## How has this been tested?

New unit test checking orchestration.

## Checklist

### DIFF
--- a/ndsl/dsl/dace/orchestration.py
+++ b/ndsl/dsl/dace/orchestration.py
@@ -551,7 +551,9 @@ def orchestrate(
         func.__annotations__[argument] = DaceCompiletime
 
     for arg_name, annotation in func.__annotations__.items():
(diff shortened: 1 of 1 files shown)


romanc (approved):
Nice!

PR #301, opened 2025-10-31
column min/max stencil - value and index

CharlesKrop:
# Description

Added two stencils to the NDSL global repository: column_max and column_min, which look at a column (or subset of a column) and identify both the max/min value and the index of the max/min value. This pattern has been observed regularly in science code but has thus far been lacking a proper mechanic in NDSL. While not a truly new feature, this PR introduces two new functions which users can call with easy.

## How has this been tested?

Added a test which checks both functions.

### DIFF
--- a/ndsl/stencils/column_operations.py
+++ b/ndsl/stencils/column_operations.py
@@ -1,0 +1,49 @@
+from ndsl.dsl.gt4py import function
+
+
+@function
+def column_max(field, start_index, end_index):
+    """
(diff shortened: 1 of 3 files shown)


FlorianDeconinck (changes requested):
Looking good.

Few comments and also should we make a stencil version for `FloatField`? As a "common case" helper?

FlorianDeconinck (approved):

PR #304, opened 2025-11-03
[Feature] Schedule Tree:  refine transient

FlorianDeconinck:
This PR is mostly to record the basics of a pass to achieve better cache access.

The refine transient optimization looks to reduce the dimensionality of transients as much as possible.

Originally we were going to use the `dace.SDFG` transform `RefinedNestedAccess` but we can't get it to work properly with our small SDFG.

This first pass work on a very small subsets of cases:
- no data flow analysis
- works for uni-dimensional maps (pass is deployed to insure that)
- no checks for conditionals or other `memlet` fixing
- ... (probably more)

Also:
- Better pipeline to apply schedule tree optimization passes
- Fix axis merge pass
- Unit tests: fix previous one and add something for refine transient

### DIFF
--- a/ndsl/dsl/dace/orchestration.py
+++ b/ndsl/dsl/dace/orchestration.py
@@ -11,7 +11,6 @@ from dace import dtypes
 from dace import method as dace_method
 from dace import nodes
 from dace import program as dace_program
-from dace.dtypes import AllocationLifetime
 from dace.dtypes import DeviceType as DaceDeviceType
 from dace.dtypes import StorageType as DaceStorageType
 from dace.frontend.python.common import SDFGConvertible


ndsl/dsl/dace/stree/optimizations/refine_transients.py
```diff
@@ -0,0 +1,150 @@
...
+                )
+
+    def visit_ScheduleTreeRoot(self, node: stree.ScheduleTreeRoot) -> None:
+        self.containers = node.containers
+        for child in node.children:
+            self.visit(child)
+
+
+class CartesianRefineTransients(stree.ScheduleNodeTransformer):
+    """ """
```
romanc: I'd either add a comment or remove the empty one

ndsl/dsl/dace/orchestration.py
```diff
@@ -157,16 +160,44 @@ def _build_sdfg(
...
+                                CartesianAxisMerge(AxisIterator._J),
+                                CartesianRefineTransients((1, 0, 2)),
+                            ]
+                        )
+                    CPUPipeline(passes=passes).run(stree)
 
             with DaCeProgress(config, "Schedule Tree: go back to SDFG"):
+                for _, data in stree.get_root().containers.items():
+                    if data.transient:
+                        data.lifetime = AllocationLifetime.State
```
romanc: what's the purpose of this change in lifetime? are you sure this is okay to do in general? if so, can we make it part of the gt4py-dace bridge?

tests/stree_optimizer/test_optimization.py
```diff
@@ -56,14 +63,63 @@ def test_stree_roundtrip_no_opt() -> None:
...
+        super().__init__(stencil_factory.config.dace_config)
+        self.stencil_A = stencil_factory.from_dims_halo(
+            func=stencil_A,
+            compute_dims=[X_DIM, Y_DIM, Z_DIM],
+        )
+        self.tmp = self.make_local(quantity_factory, [X_DIM, Y_DIM, Z_DIM])
+
+    def __call__(self, in_field: Quantity, out_field: Quantity) -> None:
+        self.stencil_A(in_field, self.tmp)
+        self.stencil_A(self.tmp, out_field)
```
twicki: The fact that this works is so cool! We open up a world of options with true cross-stencil optimization at that level

ndsl/dsl/dace/stree/optimizations/refine_transients.py
```diff
@@ -0,0 +1,170 @@
...
+import dace.sdfg.analysis.schedule_tree.treenodes as stree
+
+from ndsl import ndsl_log
+from ndsl.dsl.dace.stree.optimizations.memlet_helpers import AxisIterator
+
+
+def _zero_index_of_tuple(tuple_: tuple[int, ...], index: int) -> tuple[int, ...]:
+    new_list = list(tuple_)
+    new_list[index] = 1
+    return tuple(new_list)
```
romanc:
```suggestion
def _change_index_of_tuple(old_tuple: tuple[int, ...], index: int, value: int = 1) -> tuple[int, ...]:
    """Return a copy of the given tuple with `old_tuple[index]` being replaced by `value`.
    
    Args:
        old_tuple: to be copied
        index: at which index to replace a value
        value: to replace `old_tuple[index]`
    """
    new_list = list(old_tuple)
    new_list[index] = value
    return tuple(new_list)
```

ndsl/dsl/dace/stree/optimizations/refine_transients.py
```diff
@@ -0,0 +1,170 @@
...
+        if axis_iterator.value[0] in map_entry.params[0]:
+            access_in_map_count += 1
+
+    if access_in_map_count != 1:
+        return False
+
+    # If this transient is used in exactly one single-Axis map
+    # therefore this dimension can be removed. BUT we are not truly
+    # removing out, we are reducing it to 1 to not have to deal
+    # with different slicing
```
romanc:
```suggestion
    # This transient is used in exactly one single-Axis map
    # therefore this dimension can be removed. BUT we are not truly
    # removing it, we are reducing it to 1 to not have to deal
    # with different slicing.
```

romanc (approved):
:rocket:

FlorianDeconinck:
> > @romanc RE `Scope` vs `State`.
> > Digging in I can now understand what's going on. True `transients` like we are able to flag them now via `Local` come in as `Scope`. `make_transients_persistent` won't apply to those because it figures `Scope` is a heap allocation and that's more optimal (fair?).
> > So we have not a full control over what we pool. I am going to remove the code for now - this is a bigger subject than just this.
> 
> 'Scope' is the default 'AllocationLifetime' and 'Locals' are parsed and created by the parser if DaCe. They don't go through 'our' translation layer because they aren't allocated within stencils. So that part makes sense. If we want to put 'Loxal's into the pool, we can probably write a simple pass on the stree to re-scope them. Ii agree, this sounds like a bigger / different discussion.

Exactly. Because if you are allocating say a 1D array of 137 level on the heap because we merged well: I want it on the heap. But if we fail merging and you are going full 3D on the heap, huh, yeah I want it pooled!

And we can even go deeper looking at cache size and have a L2 restriction for those.

So yeah, future PR.

PR #307, opened 2025-11-04
refactor: directly use gt_storage in `QuantityFactory`

romanc:
# Description

Tracer bundles got me looking how we allocate and size data ... this is fallout. I realized that after #228, we don't need the extra allocation class anymore in `QuantityFactory`. The class was only there to "normalize" gt4py storage object to something that is comparable to numpy/cupy allocation.

Related to issue #230 (like not quite there - just a step in the right direction).

## How has this been tested?

Assumed to be covered by existing unit tests. We just remove an (empty) layer of indirection here.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas: N/A
- [ ] I have made corresponding changes to the documentation (e.g. add new modules to docs/docstrings/): N/A
- [x] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published in downstream modules: N/A
- [ ] New check tests, if applicable, are included: N/A

Issue #230, opened 2025-09-17
refactor: Quantities and data allocation


romanc:
# Description

This issue outlines a refactor of the data allocation behind `Quantity` objects. (Physical) Quantities are the objects in NDSL that hold data allocated in performance optimized way. They are ideally generated by a `QuantityFactory` (which is available from the `boilerplate` module).

## Current limitations

Discussions in PR https://github.com/NOAA-GFDL/NDSL/pull/222, triggered a series of questions not only about the `QuantityFactory` but also about the `Quantity` object itself and how much of the allocation details we should actually hide away from NDSL users.

Currently, the data allocator behind a Quantity can be any "numpy compatible" module, e.g., `numpy` or `cupy`.  The allocator can also be a `gt4py.storage` object, which - in turn - use either `numpy` or `cupy` under the hood. There isn't a "numpy-like" API an the variety of possible allocators shows through in different places: Quantity objects have an optional `gt4py_backend` in case the allcotor is based on `gt4py.storage`. Halo exchange code - on the other hand - is using more "numpy terminology" and is using plain `cupy` to write direct cuda code.

## Improvement ideas

In the future (we might need to split this into several refactors), we'd like to see

- `gt4py.storage` as the sole provider for data allocation of a `Quantity`
- a numpy-like API conserved and documented as "like" without mentioning the reality of the allocator
- device residency metadata and/or a way to reach the factory and/or a way to re-allocate on the same device

/cc @FlorianDeconinck as discussed

PR #309, opened 2025-11-05
RFC: refactor: quantity make gt4py_backend mandatory in the future

romanc:
# Description

Following up from https://github.com/NOAA-GFDL/NDSL/pull/307, this PR suggests to make `gt4py_backend` a required argument for a `Quantity` object. The gt4py backend name is used to ensure an optimal data layout, copying data into a better layout if needed. The backend name is then stored with other metadata in `quantity.metadata`.

Directly creating `Quantity` objects (like without using the `QuantityFactory`) is done mostly in tests. There's sporadic usage in the halo code and in the legacy restart system.

Related to issue #230 (like not quite there - just a step in the right direction).

Questions:

1. For me, this seems to be a step in the right direction, especially in context of #230. Is that really true? Or do we (e.g. in the halo code) fundamentally rely on the fact that we can build a `Quantity` without an associated backend?
2. Assuming we go ahead with this: Should we rename `gt4py_backend` -> `backend`?
3. Should we stop allowing basic data types in the `__init__` of `Quantity`? I mean, the type-hint of `data` is already set to `np.ndarray | cupy.ndarray`. Not that this would stop anyone in the python world ...
https://github.com/NOAA-GFDL/NDSL/blob/2c9946c0d74ff8d557db5ccdf8acc8e94a73d391/ndsl/quantity/quantity.py#L81-L83
4. What are `quantity.attrs`? From usage in `test_transpose.py`, I assume it's meant as a dumping ground for additional information stored on (and propagated with) quantities. If so, should we allow to set those `attrs` in the constructor and "pull them up" into the `ones()`, `zeros()`, `empty()`, ... of `QuantityFactory`? That way users don't need to do the two-step `q = quantity_factory.zeros(); q.attrs = {"my": "description here"}`

## How has this been tested?

Please describe the tests that you ran to verify your changes. Please also note
any relevant details for your test configuration (e.g. compiler, OS).  Include
enough information so someone can reproduce your tests.

## Checklist


FlorianDeconinck:
1. Halo exchange don't allocate `Quantity`. Allocations are in a buffer pool.
I am not against removing the capacity to allocate `Quantity` by hand. `QuantityFactory` are near-trivial to build, taking only a `GridSizer` and a `backend`. We can think on how to lighten even more this API.

2. We should rename everything `backend`, unless it's an internal concept which it isn't here.

3. We don't use `Quantity` for scalar, though they were designed to be transparent. Boils down to our native support for scalars, which seems relatively solid. Could make automatic `data -> Quantity` conversion a bit more tedious. Minor, would remove.

4. Those are used for interoperability with `xarray` (see `NetcdfMonitor`, `ZarrMonitor`, `test_translate.py`). I agree they are not in right level, those are metadata and should be stored there. I'd have to look again at exactly what those carry and if we need to cache them at all or if they can be re-computed on demand

PR #311, opened 2025-11-06
[Orchestration/Build]: Flag to force single cache for orchestration 

FlorianDeconinck:
# Description

Introduce a `single_code_path` flag in the DaCeConfig that forces a single cache to be built. This is the case of the doubly-periodic or tile grid, which is the case of most column-physics. It allows a single rank to compile, which in the case of many ranks compiling (HPC) dimishes the ram & disk pressure.

## How has this been tested?

Tested in GEOS - this remains an "hidden" feature which should be superseeded by a better build/cache system for orchestation.

## Checklist


romanc (approved):
LGTM

Issue #298, opened 2025-10-31
Redesign of the hierarchy of `Comm` classes


romanc:
# Description

PR #292, the subsequent issue #294 and the presence of

https://github.com/NOAA-GFDL/NDSL/blob/d8027ea376752d6c80f24ece8fa38df23063e0d3/ndsl/testing/dummy_comm.py#L1

bring up design questions about the hierarchy of `Comm` classes. This issue serves as snapshot of the current state, which will be the ground for upcoming discussions.

## Current state

```mermaid
---
title: Comm classes
---
classDiagram
    direction LR
    namespace CachingComm {
        class CachingCommReader
        class CachingCommWriter
    }
    Comm <|-- CachingCommReader
    Comm <|-- CachingCommWriter
    Comm <|-- DummyComm
    note for DummyComm "New since PR #292"
    Comm <|-- LocalComm
    note for LocalComm "Exposed in `ndsl/__init__` as `DummyComm` - used in tests"
    Comm <|-- MPIComm
    note for MPIComm "Abstraction layer for mpi4py"
    Comm <|-- NullComm
    note for NullComm "Currently used in boilerplate module"
```

Quick description of the classes

| class | description |
| --- | --- |
| CachingComm* | used in in `snapshots` (see #157) to compare the output of different versions of ported code |
| MPIComm | wrapper around the communicator of `mpi4py` (which used to be an optional dependency) |
| LocalComm | implements `send()` / `receive()` with local buffers; can't do `allgather` and `allreduce` by design (no mpi) |
| NullComm | "local" communicator in which `receive()` is hard-coded to return a fixed `fill_value` (no mpi) |
| DummyComm | unfinished single-rank communicator (no mpi) |

## Observations

- `CachingComm*` classes are linked to the "snapshot system", which has an unclear future, see #157 
- There seems to be redundancy between `LocalComm`,  `NullComm`, and the `DummyComm` from the `ZarrMonitor`
- There is no clear guidance when to use which communicator.
- In the top-level `ndsl/__init__.py`, `LocalComm` is exposed under two names: `LocalComm` and `DummyComm`. This `DummyComm` is used in tests.

## Suggested tasks

In this section, I've already put some suggestion based on my analysis of the current state. All of the suggestions are - as the name says - suggestions. This list is obviously work in progress and subject to change. I'd like to change/amend/refine this list during/after the upcoming discussion.

[...]

Issue #178, opened 2025-07-03
[Cleanup] Use proper `pytest.marker` to skip parallel test in serial


FlorianDeconinck:
(per @romanc)

Just to disambiguate: Some tests (the ones inside tests/mpi) rely on MPI being changed to None in case of a serial run:

[NDSL/tests/mpi/mpi_comm.py](https://github.com/NOAA-GFDL/NDSL/blob/2986b450386b5006d847f246ff6e8b23abdc9190/tests/mpi/mpi_comm.py#L1-L6)
```python
 from mpi4py import MPI 
  
  
 if MPI.COMM_WORLD.Get_size() == 1: 
     # not run as a parallel test, disable MPI tests 
     MPI = None 
```

Tests outside test/mpi import MPI from mpi4py directly and thus check the world size themselves. Might be worth cleaning this up anyway ... or - you know - use pytest.marker instead. Feel free to leave this open if you want to discuss this.

FlorianDeconinck:
@romanc confirmed done (looks like it) and close please

romanc:
It's done except in translate tests. Translate tests have other issues, so I think it's okay to close this issue. If we really want, we can open one for the translate tests, but I wouldn't know where to start and I think we won't have capacity anyway.

FlorianDeconinck:
>  but I wouldn't know where to start and I think we won't have capacity anyway.

Agreed. I think in the case of translate tests the answer is "when/if we rewrite the entire system..."

PR #318, opened 2025-11-12
refactor: Deprecate `NullComm` in favor of `MPIComm` and `LocalComm`

romanc:
# Description

As discussed in issue #298, we'd like to deprecate `NullComm` in favor of either `MPIComm` or `LocalComm`. This PR removes usage of `NullComm` from the NDSL codebase and adds the `DeprecationWarning`.

Notable / non-trivial changes:

- boilerplate now uses `MPIComm` and complains if users try to initialize single-tile factories with an MPI size > 1.
- tests `test_set_hybrid_pressure_coefficients_correct` and `test_unblock_waiting_tiles` moved to the `tests/mpi` folder because they rely on asynchronous send/receive, which `LocalComm` apparently can't really handle. If I try to use it, I end up with an error claiming "receive was called before send".
- two tests in `tests/test_caching_comm.py` needed changes (because `NullComm` was able to receive without sending anything (because sending was hard-coded, so you wouldn't need to send))
- tests under `tests/mpi` now import their version of `MPI` from `tests/mpi/__init__.py` rather than `tests/mpi/test_mpi.py` (which didn't contain any tests). Issue #178 will eventually fix this "the right way"

## How has this been tested?

Covered by existing tests.

## Checklist

### DIFF
--- a/ndsl/comm/null_comm.py
+++ b/ndsl/comm/null_comm.py
@@ -1,4 +1,5 @@
 import copy
+import warnings
 from collections.abc import Mapping
 from typing import Any, TypeVar, cast
 
@@ -33,6 +34,12 @@ class NullComm(Comm[T]):
             fill_value: fill halos with this value when performing
                 halo updates.
         """
+        warnings.warn(
+            "NullComm is deprecated and will be removed with the next version of NDSL. "
+            "Use MPIComm or LocalComm instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
         self.rank = rank
         self.total_ranks = total_ranks
         self._fill_value = fill_value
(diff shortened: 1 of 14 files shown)


romanc:
> > rely on asynchronous send/receive, which LocalComm apparently can't really handle. If I try to use it, I end up with an error claiming "receive was called before send"
> 
> We should make sure of this. It's "implemented", e.g. IRecv/ISend are there, but if it doesn't work we should error out.

Okay, let me see if I can make it work. Otherwise, I'll error out.

FlorianDeconinck (approved):
🧹 ALL SHALL BE CLEANED 🧹 

Thanks mate, another one for the books!

PR #324, opened 2025-11-14
New functions: column_min_ddim & column_max_ddim

CharlesKrop:
Functionality is the same as column_min/max, but separate functions are needed to handle cases with off grid data dimensions.

The unit test `test_column_operations` has been updated to test both 3d and 4d cases.


FlorianDeconinck (approved):
LGTM

PR #335, opened 2025-11-27
[API Update] `NDSLRuntime` uses the wider `StencilFactory` rather than a narrow config

FlorianDeconinck:
# Description

We fix here a misguided choice of original API _before_ the class becomes wide spread.

Previous API of `NDSLRuntime` took the only required argument, e.g. `DaCeConfig`. This is quite narrow and will lead any extension of the system to lack data and have an API update. On top of it, the dace config is burried, in most cases, within the stencil factory, which is not user friendly.

We prefer here to pass down the full `StencilFactory`, which is a common factory to carry around model code, and extract what we need for it. `StencilFactory` carries configs _and_ grid indexing, which should cover most basic information.

## How has this been tested?

Update to the utest

## Checklist

### DIFF
--- a/tests/dsl/orchestration/test_call.py
+++ b/tests/dsl/orchestration/test_call.py
@@ -64,7 +64,7 @@ class DefaultTypeProgram(NDSLRuntime):
         stencil_factory: StencilFactory,
         quantity_factory: QuantityFactory,
     ):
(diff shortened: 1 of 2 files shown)


twicki (approved):
Looks good, ready to go!

PR #336, opened 2025-12-08
:wrench: Harden `Quantity` to hybrid usage

FlorianDeconinck:
# Description

Bits and bobs that make using `Quantity` in the case of hybrid CPU/GPU memory more straightforward for the user:
- Check for device residency in allocation and move to upload/download if needed
- Make sure to download data before using the `xarray` getter
- Loosen the `from_data_array` to read data that doesn't have a `units`

## How has this been tested?

We don't have the possibility to do GPU tests for now so we can test this outside of model use. This has been tested in NASA's GEOS current porting work.

## Checklist

### DIFF
--- a/ndsl/quantity/quantity.py
+++ b/ndsl/quantity/quantity.py
@@ -359,19 +359,19 @@ class Quantity:
     @property
     def field_as_xarray(self) -> xr.DataArray:
         """Returns an Xarray.DataArray of the field (domain)"""
-        if isinstance(self.field, cupy.ndarray):
-            field = self.field.get()
-        else:
+        if isinstance(self.field, np.ndarray):
             field = self.field
+        else:
+            field = self.field.get()
         return xr.DataArray(field, dims=self.dims, attrs=self.attrs)
 
     @property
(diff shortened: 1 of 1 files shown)


ndsl/quantity/quantity.py
```diff
@@ -347,12 +359,20 @@ def extent(self) -> tuple[int, ...]:
     @property
     def field_as_xarray(self) -> xr.DataArray:
         """Returns an Xarray.DataArray of the field (domain)"""
-        return xr.DataArray(self.field, dims=self.dims, attrs=self.attrs)
+        if isinstance(self.field, np.ndarray):
+            field = self.field
+        else:
+            field = self.field.get()
```
twicki: Do you know how this works under the hood? It is definetly the right thing to do but I'm curious if transporting bounded array views from device to host is slow, as fast as copying the full thing or even faster? Because if slicing it was really slow, we might consider a `data.get()` followed by a cpu-slicing. But that is only if we see this being an issue

the xarray stuff is not on the critical path, so we can go either way, but I'm just curious.

PR #340, opened 2025-12-11
[Feature] `LocalState` allocation as `State` for test

FlorianDeconinck:
We allow the the `LocalState` to be allocated as a regular `State` by swapping all of it's element to `Local`. This is useful for testing and interfacing code when the user wants to re-use an internal State of higher numerics in runtime order.

## How has this been tested?

New unit tests + used in GEOS GFDL 1M port.

## Checklist

- [x ] My code follows the style guidelines of this project

### DIFF
--- a/tests/quantity/test_local.py
+++ b/tests/quantity/test_local.py
@@ -149,3 +149,19 @@ def test_forbidden_access_to_locals() -> None:
         match="State contains Quantity my_quantity but is allocated as a LocalState. LocalState with Locals can _only_ contain Locals.",
     ):
         the_code.allocate_bad_locals()
+
+
+def test_local_state_as_regular_state() -> None:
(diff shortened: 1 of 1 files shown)


FlorianDeconinck:
@oelbert  / @CharlesKrop 

This solves the testing problem where you have

```
Class A
        LocalStateA
        Class B
                 ...
```

And the Class B call relies on some of `LocalStateA`. You want to be able to write your test of ClassB

```python

locals_ = ClassA.LocalStateA.make_locals()
B(locals.xx, ...)
```

But that's using `Locals` outside of there defining class, which is forbidden. So you use `make_as_state` and you get a proper `State`, which won't yell.

Of course, if this is use at proper runtime, all optimization will be destroyed (because no Locals) and I will be very sad.

PR #345, opened 2025-12-17
[Feature] Introducing `Xumpy`

FlorianDeconinck:
# Description

One of the issue with performance portability for numerics in Python is the reliance on `numpy` for speedier CPU numerics. The `cupy` packages offers a mirroring API for CUDA/ROCm portability. There's no true unified package as of today.

Atop of this, we carry some of our own concepts  (`Quanitty`, package-wide literal precision,etc.).

We propose here to introduce `xumpy` (name can be changed) a unified API for our users to write numerics numpy-like we the DSL,

Development of this sub-package is to be done on "need to use" basis as it will be easy to add features.

**Naming**: I chose `xumpy` to have `x` as the "variable" that brings `numpy` and `cupy` together. Another name could be `pumpy` with "p" for "performance portable". I am happy to hear _all_ of your good and less good ideas, so we may solve The Naming Problem sooner rather than later.

## How has this been tested?

New unit tests, deployed in GEOS's GFDL 1M and updated FV3 code

## Checklist

### DIFF
--- a/ndsl/xumpy/minmax.py
+++ b/ndsl/xumpy/minmax.py
@@ -7,6 +7,7 @@ from ndsl import Quantity
 from ndsl.dsl.typing import Float
 from ndsl.optional_imports import cupy as cp
 
+
 if cp is None:
     cp = np
 


fmalatino (approved):

ndsl/xumpy/minmax.py
```diff
@@ -0,0 +1,85 @@
...
+    out_buffer: npt.NDArray | Quantity,
+) -> None:
+    """Find maximum value on each horizontal plane, e.g. for the
+    first two dimensions of the array."""
+    raise NotImplementedError("")
+
+
+@max_on_horizontal_plane.register(np.ndarray | cp.ndarray)
+def _(in_buffer: npt.NDArray, out_buffer: npt.NDArray) -> None:
+    print("numpy/cupy")
```
twicki: rm?

ndsl/xumpy/minmax.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+@singledispatch
+def max_on_horizontal_plane(
+    in_buffer: npt.NDArray | Quantity,
+    out_buffer: npt.NDArray | Quantity,
+) -> None:
+    """Find maximum value on each horizontal plane, e.g. for the
+    first two dimensions of the array."""
+    raise NotImplementedError("")
```
twicki:
```suggestion
    raise NotImplementedError(f"max_on_horizontal_plane called with unsupported type {type(in_buffer)}")
```

jjuyeonkim (approved):

PR #347, opened 2025-12-29
Parameterize `SubtileGridSizer` to allow variation of the padding on non-interface field

FlorianDeconinck:
**:bricks::bricks:  BLOCKED BEHIND https://github.com/GridTools/gt4py/pull/2426 :bricks::bricks:**

# Description

In order to allow zero-copy memory mapping from Fortran (case of GEOS) we need to parameterize `GridSizer` and children to allow variation of the padding on non-interface field. This PR takes care of it by requiring the backend and checking it's Fortran friendliness. Further discussion should question the need for padding altogether.

This will lead to an API break down the road, as we know require the backend to be given to the `SubtileGridSizer`.

Also:
- Add function to check that backend is "fortran aligned"

## How has this been tested?

New unit test.

## Checklist

### DIFF
--- a/ndsl/initialization/grid_sizer.py
+++ b/ndsl/initialization/grid_sizer.py
@@ -15,8 +15,6 @@ class GridSizer(ABC):
     """Number of horizontal halo points for produced arrays."""
     data_dimensions: dict[str, int]
     """Name/Lengths pair of any non-x/y/z dimensions, such as land or radiation dimensions."""
-    internal_pad_non_interface_dimensions: bool = False
-    """Enforce interface and non-interface dimensions true memory to be the same"""
 
     @abstractmethod
     def get_origin(self, dims: Sequence[str]) -> tuple[int, ...]: ...


FlorianDeconinck:
@oelbert for visibility.

In Pace after the January release, you will be required to give `backend` to the `SubtileGridSizer`. For now nothing change but warning will pop.

ndsl/initialization/grid_sizer.py
```diff
@@ -15,6 +15,8 @@ class GridSizer(ABC):
     """Number of horizontal halo points for produced arrays."""
     data_dimensions: dict[str, int]
     """Name/Lengths pair of any non-x/y/z dimensions, such as land or radiation dimensions."""
+    internal_pad_non_interface_dimensions: bool = False
+    """Enforce interface and non-interface dimensions true memory to be the same"""
```
romanc: As far as I can see, `internal_pad_non_interface_dimensions` is internal to `SubtileGridSizer`, right?

Assuming so, I wouldn't change the `GridSizer` base class and "hide" this `SubtileGridSizer`. One way I could see this happen is by adding an `__init__()` to `SubtileGridSizer`, which would look something like

```py
class SubtileGridSizer(GridSizer):
    def __init__(nx, ny, nz, n_halo, data_dimensions, backend = None):
        super(nx, ny, nz, n_halo, data_dimensions)

        if backend is None:
            warnings.warn(
                ...
            )
            self._pad_non_interface_dimensions = True
        else:
            fortran_style_memory = backend_is_fortran_aligned(backend)
            self._pad_non_interface_dimensions = not fortran_style_memory
...
```

That way, we

1. do not compromised the idea that `GridSizer`s should - theoretically - be independent of backends
2. found a way to make an exception for a specific subclass, namely for `SubtileGridSizer` to be backend dependent
3. and we keep the (possible) allocation of an extra interface point local to `SubtileGridSizer`

romanc (changes requested):
If we can avoid it, I wouldn't change the `GridSizer` base class (see inline).

ndsl/initialization/subtile_grid_sizer.py
twicki: just generally if we're already touching this: Is there a case where we want to turn on or off `(ij)` and `k` padding independently - physics vs dynamics? Or do we acutally think padding shoud in the long term go completely and not worry about it?

ndsl/dsl/gt4py_utils.py
```diff
@@ -450,6 +450,23 @@ def is_gpu_backend(backend: str) -> bool:
     return gt_backend.from_name(backend).storage_info["device"] == "gpu"
 
 
+_FORTRAN_LOOP_LAYOUT = (2, 1, 0)
+"""Fortran is a column-first (or stride-first) memory system,
+which in the internal gt4py loop layout means I (or axis[0]) has
+the higher value, e.g. "higher importance to run first"""
```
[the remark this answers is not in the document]
FlorianDeconinck: Done

romanc (commented):
Nice :+1:

PR #348, opened 2025-12-29
[Feature | Orchestration] Optimizated C call overhead away, update pipeline, optimize CPU transient residency

FlorianDeconinck:
# Description

DaCe orchestration was plagued with a slow integration due to the routine marshalling python object (arrays as arguments but also closure) into C-binding ready pointers for calling into the C library

We originally cached _everything_ at first call but that lead to instability: any re-allocation, different argument made to the same program _silently_ fail and we reverted it.

This PR introduces a proper argument hashing that reduces the overhead to being negligible impact on runtime while keeping stability for changing arguments. The hypothesis goes as follows:
- Closure _do not change_ because we build _and_ run or we expect pipeline to have taken care of this
- Arguments are either NDSL owned (Quantity, State) or __array_interface__ compatible (no DLPack yet)

This PR also updates the orchestration pipeline:
- prepare for proper use of the soon to be merge `dace:cpu_KJI` (https://github.com/GridTools/gt4py/pull/2426)
- stree optimizer (still experimental) covers all backends
- CPU transient memory is NOT pushed into pool memory (because of a lack of pooling for CPU)
- GPU optimization run the complete `dace` auto-optimizer by default

## How has this been tested?

Unit tests and on the microphysics benchmark conducted with GEOS.

## Checklist

### DIFF
--- a/ndsl/dsl/dace/dace_executable.py
+++ b/ndsl/dsl/dace/dace_executable.py
@@ -10,7 +10,7 @@ from ndsl.quantity import State
 
 @dataclasses.dataclass
 class DaceExecutable:
-    """Bundle the executable (lib) and it's marshalled
+    """Bundle the executable (lib) and its marshalled
     arguments for execution"""
 
     compiled_sdfg: dace.CompiledSDFG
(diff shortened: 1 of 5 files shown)


FlorianDeconinck:
Bringing back to draft: the hashing system operates under the assumption the types given to are the orchestrated code are NDSL's OR trivially hashable. I'll introduce a system that deactivates the hashing if we encounter none of those and warns once of performance degradation

fmalatino (approved):

PR #349, opened 2026-01-05
Minor fix to column min/max functions

CharlesKrop:
Modified the starting min/max index of column min/max functions to start_index instead of 0. This fixed a potential issue where arrays with absolute maximum/minium occuring before the target window (i.e. array had absolute maximum at index 2 but function was called to find the max in indexes [5:10]) would return the wrong value.

Also changed column min/max test to test for this possibility

This did not fix an existing issue, but prevented a potential one.


FlorianDeconinck (approved):
Good catch. 

(which probably was downstream of an horrible debug)

PR #361, opened 2026-01-14
refactor: Types for non-deprecated corner code

romanc:
# Description

This PR adds types for the non-deprecated corner code. This allows to remove the current exception for `ndsl.stencils.corners` in the `mypy` configuration. Also removed exceptions for previously deprecated and now deleted files.

## How has this been tested?

All good as long as `mypy` is happy :)

## Checklist


ndsl/stencils/corners.py
```diff
@@ -7,6 +9,10 @@
 from ndsl.dsl.typing import FloatField
 
 
+FillCornersDirection: TypeAlias = Literal["x", "y"]
+GridType: TypeAlias = Literal["A", "B", "C", "D", "E"]
```
twicki: future proof it by supporting more than A and B here since these are the two that are supported here?

ndsl/stencils/corners.py
```diff
@@ -762,7 +874,7 @@ def fill_corners_dgrid_defn(
     y_in: FloatField,
     y_out: FloatField,
     mysign: float,
-):
+) -> None:
```
twicki: I would not add return value to stencils

twicki (changes requested):

ndsl/stencils/corners.py
```diff
@@ -59,18 +65,18 @@ def fill_corners_2cells_mult_x(
...
 
-def fill_corners_2cells_y_stencil(q_out: FloatField, q_in: FloatField):
+def fill_corners_2cells_y_stencil(q_out: FloatField, q_in: FloatField) -> None:
     with computation(PARALLEL), interval(...):
         q_out = fill_corners_2cells_mult_y(q_out, q_in, 1.0, 1.0, 1.0, 1.0)
 
 
 @gtscript.function
-def fill_corners_2cells_x(q: FloatField):
+def fill_corners_2cells_x(q: FloatField) -> FloatField:
```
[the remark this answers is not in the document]
romanc: Good catch. We previously agreed that types on `gtscript.functions` (also for inputs) are best-case useless and worst-case misleading in the sense that they could give a false sense of security to users. In `column_operations` we've started to use no types in combination with `@typing.no_type_check`. I've propagated that pattern.

PR #398, opened 2026-03-10
Improved translate test logging

CharlesKrop:
Fixed a bug that was incorrectly computing the number of changing points (points that differed between the reference in and out files).

Added a few lines that compute the number of changing columns and failing columns.

Updated the translate test print to show report these new values, and ensure that all are shown on the by the pytest short summary.

### DIFF
--- a/ndsl/testing/comparison.py
+++ b/ndsl/testing/comparison.py
@@ -368,7 +368,7 @@ class MultiModalFloatMetric(BaseMetric):
             )
             report_local_failures = f"changing grid points: {bad_indices_count}/{self.number_changing_values} - {failures_of_changing_gridpoint_pct}%; changing columns: {bad_column_count}/{total_column_count} - {bad_column_pct}%; all grid points: {bad_indices_count}/{full_count} - {failures_of_all_grid_points_pct}%\n"
         else:
-            report_local_failures = "Something went wrong when calculating statistics one line print, see log for raw data.\n"
+            report_local_failures = f"all grid points: {bad_indices_count}/{full_count} - {failures_of_all_grid_points_pct}%\n"
         report = [
             f"{report_local_failures}"
             f"Index   Computed   Reference   "
--- a/ndsl/testing/comparison.py
+++ b/ndsl/testing/comparison.py
@@ -260,11 +260,11 @@ class MultiModalFloatMetric(BaseMetric):
         # We might have sliced outputs in the translate test. Rather than funnel the slicing
         # all the way down, we bail out if we can measure input vs reference
         if input_values is not None and input_values.shape == reference_values.shape:
-            self.number_changing_values = (
-                (input_values != reference_values).sum()
-            )
+            self.number_changing_values = (input_values != reference_values).sum()
             if len(input_values.shape) == 3:
-                self.changing_column_map = (input_values != reference_values).any(axis=2)
+                self.changing_column_map = (input_values != reference_values).any(
+                    axis=2
+                )
             else:
                 self.changing_column_map = np.nan
         else:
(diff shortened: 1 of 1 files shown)


ndsl/testing/comparison.py
```diff
@@ -260,11 +260,16 @@ def __init__(
         # We might have sliced outputs in the translate test. Rather than funnel the slicing
         # all the way down, we bail out if we can measure input vs reference
         if input_values is not None and input_values.shape == reference_values.shape:
-            self.number_changing_values = (
-                (input_values != reference_values).flatten().shape[0]
-            )
+            self.number_changing_values = (input_values != reference_values).sum()
+            if len(input_values.shape) == 3:
```
FlorianDeconinck, line 265: `>=3` ? Technically any data dimensions is just another dimension on top of the grid and we could report column per tracers/radiation bands that way.,

ndsl/testing/comparison.py
```diff
@@ -260,11 +260,16 @@ def __init__(
...
+                self.changing_column_map = (input_values != reference_values).any(
+                    axis=2
+                )
+            else:
+                self.changing_column_map = np.nan
```
FlorianDeconinck: `None`?

ndsl/testing/comparison.py
```diff
@@ -346,12 +366,11 @@ def report(self, file_path: str | None = None) -> list[str]:
             failures_of_changing_gridpoint_pct = round(
                 100.0 * (bad_indices_count / self.number_changing_values), 2
             )
-            report_local_failures = f"Failures (changing grid points) ({bad_indices_count}/{self.number_changing_values}) ({failures_of_changing_gridpoint_pct}%)\n"
+            report_local_failures = f"changing grid points: {bad_indices_count}/{self.number_changing_values} - {failures_of_changing_gridpoint_pct}%; changing columns: {bad_column_count}/{total_column_count} - {bad_column_pct}%; all grid points: {bad_indices_count}/{full_count} - {failures_of_all_grid_points_pct}%\n"
```
FlorianDeconinck: Can we shorten this since it's a one liner
```
Failures (points #, column #, grid #): #,% |  #,% |  #,%
```
or something?

ndsl/testing/comparison.py
```diff
@@ -338,6 +343,21 @@ def report(self, file_path: str | None = None) -> list[str]:
...
         bad_indices_count = len(failed_indices[0])
+        if self.changing_column_map is not None:
+            if self.success.ndim == 3:
+                bad_column_count = (
+                    np.logical_not(self.success).any(axis=2) & self.changing_column_map
+                ).sum()
+                total_column_count = self.changing_column_map.sum()
+                bad_column_pct = round(bad_column_count / total_column_count * 100, 2)
+            else:
+                bad_column_count = np.nan
```
FlorianDeconinck: I think we want to prefer `None` rather than `nan` which would point to an actual computation. The use of `NaN` as a neutral value by the domain scientists is not a good idea

FlorianDeconinck (changes requested):
Looking good - maybe try a few different messages to shorten it? This is an expert system, we don't necessarily needs to be as verbose  as I made it in the past

FlorianDeconinck:
Another thought: if "changed" grid points == "all", maybe print only of?

ndsl/testing/comparison.py
CharlesKrop: this is true, but I don't have a good test case at the moment. I know I could make one, but I don't have time right now. I would rather get this in for 3d and come back to it when we need it for 4d, but if you want to wait I can figure it out next week

CharlesKrop: I also think 4d might be a bit much for one line prints anyway. if we have 23 tracers at c24 (assuming 50% are changing in the code) then saying you failed X/6500 columns isn't particularly useful. IMO it needs to be broken down by tracer, either for columns or you give a "failing tracer" count instead of failing column.

either way, I think its something that requires more thought and is deserving of its own feature/PR.

we've also never had a desire for this hypothetical feature I just dreamed up, I would argue because 1. tracers are rare and 2. when we do hit them its easy enough to just open up the netcdf and flip through since it doesn't happen particularly often

all that said I'm happy to come back, think more about it, and build it when we find a need for it

CharlesKrop:
I think there is value to printing both changing points and all points, even when they are the same. It hardly adds any clutter (in fact, I would argue keeping the print the same for all cases makes it easier to read) and  it makes it easy to recognize that all points actually change.

particularly useful since - when there is no input data to compute changing points, I only print failures of all grid points. I want to keep those two prints fairly distinct so they can be distinguished

ndsl/testing/comparison.py
FlorianDeconinck: Sounds good - please add justification in a comment in code

FlorianDeconinck (approved):
Lint + small comment to verobose intent then you can merge

PR #402, opened 2026-03-18
Improved data_loader

CharlesKrop:
Tiny PR adding ability for data_loader to auto-detect current savepoint index in the netcdf. Needed for some newer translate tests where the existing translate test systems fail to load 4D variables. No changes to existing code, previous uses of data_loader should be unaffected.

### DIFF
--- a/ndsl/stencils/testing/savepoint.py
+++ b/ndsl/stencils/testing/savepoint.py
@@ -25,19 +25,19 @@ class DataLoader:
     def __init__(self, rank: int, data_path: Path, i_call: int) -> None:
         self._data_path = data_path
         self._rank = rank
-        self.i_call = i_call
+        self._i_call = i_call
 
     def load(
         self,
         name: str,
         postfix: str = "",
-        i_call: int = 0,
+        use_dynamic_i_call: bool = False,
     ) -> dict[str, np.ndarray | float | int]:
+        call_index = self._i_call if use_dynamic_i_call else 0
         return dataset_to_dict(
             xr.open_dataset(self._data_path / f"{name}{postfix}.nc")
             .isel(rank=self._rank)
-            .isel(savepoint=i_call)
-        )
+            .isel(savepoint=call_index))
 
 
 class Translate(Protocol):
--- a/ndsl/stencils/testing/savepoint.py
+++ b/ndsl/stencils/testing/savepoint.py
@@ -22,9 +22,10 @@ def _process_if_scalar(value: np.ndarray) -> np.ndarray | float | int:
 
 
 class DataLoader:
-    def __init__(self, rank: int, data_path: Path) -> None:
+    def __init__(self, rank: int, data_path: Path, i_call: int) -> None:
         self._data_path = data_path
         self._rank = rank
+        self.i_call = i_call
 
     def load(
         self,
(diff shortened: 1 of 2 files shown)


ndsl/stencils/testing/savepoint.py
FlorianDeconinck: Remove `i_call` in parameters and use `self.i_call`

FlorianDeconinck (changes requested):

ndsl/stencils/testing/savepoint.py
CharlesKrop: if I do that it breaks loads where there aren't save points (e.g. constant files)

CharlesKrop: right now it can be used two ways, depending on the shape of the netcdf. perhaps not ideal, but it does work

```
        self.constants = data_loader.load("GFDL_1M-constants")
        
        # DEBUG
        self.debug_input = data_loader.load("GFDL_1M_WarmRain-In", i_call=data_loader.i_call)

Comment (FlorianDeconinck) [reply ndsl/stencils/testing/savepoint.py]: Meh.

Make it an override then.

```python
    def load(
        self,
        name: str,
        postfix: str = "",
        i_call: int | None = None,
    ) -> dict[str, np.ndarray | float | int]:
        call_index = self.i_call if i_call is None else i_call
        return dataset_to_dict(
            xr.open_dataset(self._data_path / f"{name}{postfix}.nc")
            .isel(rank=self._rank)
            .isel(savepoint=call_index)
```

CharlesKrop: still breaks. i_call needs to be 0 for the constant files to work properly

This is effectively the same as your first suggestion - in both cases I will have to go through and change every previous case of `.load` and add `i_call=0`, which is what I was trying to avoid

CharlesKrop: This is the other possibility I have which ensures that old code still works. flag name is the first thing that came to mind

```py
    def load(
        self,
        name: str,
        postfix: str = "",
        use_dynamic_i_call: bool = False,
    ) -> dict[str, np.ndarray | float | int]:
        call_index = self.i_call if use_dynamic_i_call else 0
        return dataset_to_dict(
            xr.open_dataset(self._data_path / f"{name}{postfix}.nc")
            .isel(rank=self._rank)
            .isel(savepoint=call_index))

Comment (romanc) [reply ndsl/stencils/testing/savepoint.py]: what if you added a default for `i_call` in the constructor of `DataLoader`, i.e.

```diff
- def __init__(self, rank: int, data_path: Path, i_call: int) -> None:
+ def __init__(self, rank: int, data_path: Path, i_call: int = 0) -> None:
```

would that allow to remove the `i_call` parameter from the `load()` function (using `self.i_call` inside)?

romanc (commented):
I'm not sure I understand the use-cases: can you add a use-case that load the wrong data without this change?

Do you know if there any current usage of the `isel` parameter of `load`, e.g.`load(name, prefix="...", isel=5)`? I'm asking because I think this case won't be possible anymore with the currently proposed changes.

Ps, linting seems to be failing ...

FlorianDeconinck (approved):
👍  Fix lint and merge in

PR #404, opened 2026-03-23
typing: add types to `ndsl.stencils.testing.grid`

romanc:
# Description

Add type hints to the `Grid` class used in translate tests. This is not perfect, but at least we have now most types. Lots of magic happening here and if we don't have minimal type hint and type checking we'll missing things like the `Quantity` update in the first commit.

## How has this been tested?

All good as long as CI is still green. Locally, `mypy` is happy.

## Checklist

### DIFF
--- a/ndsl/stencils/testing/grid.py
+++ b/ndsl/stencils/testing/grid.py
@@ -1,7 +1,10 @@
-# type: ignore
+from typing import Sequence
 
 import numpy as np
+from f90nml import Namelist
 
+from ndsl import GridSizer
+from ndsl.comm.communicator import Communicator
 from ndsl.comm.partitioner import TilePartitioner
 from ndsl.config import Backend
 from ndsl.constants import I_DIM, J_DIM, K_DIM, N_HALO_DEFAULT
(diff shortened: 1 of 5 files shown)


ndsl/stencils/testing/grid.py
```diff
@@ -164,63 +191,32 @@ def quantity_factory(self) -> QuantityFactory:
...
-
-    def quantity_wrap(
-        self,
-        data,
-        dims=(I_DIM, J_DIM, K_DIM),
-        units="unknown",
-    ):
-        origin = self.sizer.get_origin(dims)
-        extent = self.sizer.get_extent(dims)
-        return Quantity(data, dims=dims, units=units, origin=origin, extent=extent)
```
romanc, line 200: Must be dead code because the `Quantity` constructor has a required `backend` argument. If we were still using these code paths, they'd be getting `TypeError`s.

FlorianDeconinck: Dead code for sure. Probably lined to translate system at some point to move toward a full quantity.

FlorianDeconinck (approved):
🧹 

Having a `stencils.testing.grid` has been a vexing issue for a while but I am yet to take the time to nuke this monstrosity from orbit. I am yet to understand why we can't centralize the concept of grid and have this thing kicking around... Work for another time (sic).

romanc:
yeah, I had a quick look why this class exists in the first place and very quickly stopped looking :see_no_evil: Indeed it seems this has to be redundant (or at least just for convenience), but it seems to pull in a couple of concepts that I'm not yet super familiar with, so I just left it there. To be revisited later(tm)

PR #414, opened 2026-03-31
[Config | DaCe] Return to baseline of _no_ unrolling loop in DaCe

FlorianDeconinck:
# Description

[Schema for DaCe loop unrolling](https://spcldace.readthedocs.io/en/latest/source/config_schema.html#envvar-frontend.unroll_threshold) is setting the level of unrolling as an integer, with 0 meaning "all" and -1 "None".

An error has gotten into the system where we set the value to `False` - which is nonsensical.

Furthermore, I believe the default should be to _not_ unroll. While potentially removing some capacity of code inline & helping the compiler to pre-fetch, I believe those are minimal compared to the potentially catastrophic cost in build time (loop on tracers...).

This PR reset the flag to -1, the default, and comments on the why in code.

Also (due to lint waking up on it)

- Refactor the global debugger setup so `lint` can realize that it is not _always_ `None` but can actually carries a value. Also cleaner.

## How has this been tested?

N/A


FlorianDeconinck:
Well this backfired spectacularly... A new bug for the pile!

FlorianDeconinck:
This explodes with or without the change. I have _no idea_ how any of the CI before passed...

romanc:
Linking related [DaCe issue](https://github.com/spcl/dace/issues/2332).

FlorianDeconinck:
This is ready for merge

twicki (approved):
Looks ready to go

PR #416, opened 2026-04-01
[Feature] Data Dimension Fields

FlorianDeconinck:
# Description

The `gt4py` frontend requires the data dimensions to be statically defined as part of the type for the stencil definition. This means that all dimension sizes need to be known when _importing_ code. This is impractical when:
- integrating in a larger model where we only partially control configuration ingestion
- when some initial computation is required to determine the exact size of a dimensions (case of the Convective Tracers in GEOS)

We had introduced a `FieldBundle` to fix this issue, but encountered a series of issues with orchestration.

Building on the same idea of deferred types (see `internal.deferred_type` and `dsl.stencil`), we introduce `DataDimensionFields` that expose a  `declare` (static typing) and a `register` (dynamic sizing) function, solving the above, with a new series of simpler but orchestration worthy functions to access indexation:

- `def index(name: str)`: allows to get an index of a dimension by name if it was given in the `mapping`
- `def size()`: gives the size of a dimension for loops range

## Extra code

When modifying `stencil` I ran into a `mypy` issue leading to the clean up of how the global debugger is instantiated.

## How has this been tested?

New unit tests + is developed against `pyFV3` (PR/Branch TBD)

## TODO before merging

- [ ] Remove `FieldBundle`

## Checklist


FlorianDeconinck:
@twicki / @romanc : Review of the field code please

@oelbert  for reference.

ndsl/debug/config.py
romanc, line 1: Thanks for the cleanup 🧹

ndsl/internal/deferred_type.py
```diff
@@ -0,0 +1,66 @@
+import abc
+import ast
+import inspect
+from types import FrameType
+from typing import Any, Type
+
+
+def get_lhs_name(frame: FrameType | None) -> str:
+    """Inspect the back frame to retrieve the LHS of an assign
+    operation, will fail if any of those aren't true"""
```
romanc:
```suggestion
    """Inspect the back frame to retrieve the LHS of an assign
    operation. Will fail if any of those are None."""
```

ndsl/internal/deferred_type.py
```diff
@@ -0,0 +1,66 @@
...
+    if not isinstance(target_node, ast.Name):
+        raise RuntimeError("LHS retrieval failed: AST LHS malformed")
+    return target_node.id
+
+
+class StencilTypeRegistrar:
+    """Type registrar that knows how to give back true type.
+
+    This acts as a registrar and also a dynamic resolver to be able to swap to the true
+    type before
```
romanc: before what?

ndsl/internal/deferred_type.py
```diff
@@ -0,0 +1,66 @@
...
+    This type should carry all the API of the true type given by `resolve`
+    in order to be able to work with it as if it was resolved.
+    """
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def __descriptor__(self) -> None:
+        # Ignore, use JIT
+        return None
```
romanc, line 63: I don't understand. What (type of JIT) is this descriptor for?

ndsl/quantity/data_dimensions_field.py
```diff
@@ -0,0 +1,257 @@
+import collections
+import inspect
+from dataclasses import dataclass
+from typing import Type
```
romanc: I think nowadays, lowercase `type` is generic and should work out of the box without an import.

ndsl/quantity/data_dimensions_field.py
```diff
@@ -0,0 +1,257 @@
...
+        data_dims: tuple[int] | collections.abc.Collection = tuple(),
+    ) -> None:
+        super().__init__(dtype, axes, data_dims)
+        self._mapping: SparseNameMapping = {}
+
+    def range(self, dimension_index: DataDimensionIndex) -> range:
+        return range(self.data_dims[dimension_index])
+
+    def set_mapping(self, mapping: SparseNameMapping) -> None:
+        self._mapping = mapping
```
romanc: How about we make this a proper property setter?

```suggestion
    @mapping.setter
    def mapping(self, mapping: SparseNameMapping) -> None:
        self._mapping = mapping
```

ndsl/quantity/data_dimensions_field.py
```diff
@@ -0,0 +1,257 @@
...
+        optionally a sparse mapping of name/index.
+
+        The same type cannot be registered twice and will error out.
+
+        Args:
+            pre_registration_type: Type returned by the "declare" function.
+            data_dimensions: tuple of int giving size of each data dimensions.
+            name_mapping: for each dimensions, a sparse dictionnary giving a name/index
+                to retrieve 3D fields by name.
+            dtype: Inner data type, defaults to Float.
```
romanc, line 110: Args here are out of date with args of the function

ndsl/quantity/data_dimensions_field.py
```diff
@@ -0,0 +1,257 @@
...
+        # Requires the _locals to get `name` - do not pull out of `register`
+        @oprepo.replaces(f"{name}.size")
+        def _data_dimensions_size(
+            pv: ProgramVisitor,
+            sdfg: SDFG,
+            state: SDFGState,
+            data_dim_index: int,
+        ) -> dace.int32:
+            size = cls._type_registrar[name].size(data_dim_index)
+            return dace.int32(size)
```
romanc: Should this honor literal integer precision configured in NDSL/GT4Py?

ndsl/quantity/data_dimensions_field.py
```diff
@@ -0,0 +1,257 @@
...
+        name_mapping: SparseNameMapping | None = None,
+        dtype: npt.DTypeLike = Float,
+    ) -> "DataDimensionsMarkupType":
+        """Declare a data dimension field and register it's size
+
+        Args:
+            data_dims: tuple of int giving size of each data dimensions.
+            name_mapping: for each dimensions, a sparse dictionnary giving a name/index
+                to retrieve 3D fields by name.
+            dtype: Inner data type, defaults to Float.
```
romanc: Args here are out of date with the args of the function

ndsl/quantity/data_dimensions_field.py
```diff
@@ -0,0 +1,257 @@
...
+    @classmethod
+    def get(cls, name: str) -> _DataDimensionsFieldDescriptor:
+        """
+        Declare a data dimension field - which will need to be properly
+        registered later.
+
+        Args:
+            name: name of the type as registered via `register`
+            do_markup: if name not registered, markup for a future specialization
+                at stencil call time
```
romanc:
```suggestion
        Args:
            name: name of the type as registered via `register`
```

ndsl/quantity/data_dimensions_field.py
```diff
@@ -0,0 +1,257 @@
...
+        return DataDimensionsField._type_registrar[self.name].range(dimension_index)
+
+    def _get_true_type(self) -> _DataDimensionsFieldDescriptor:
+        try:
+            return DataDimensionsField._type_registrar[self.name]
+        except KeyError:
+            raise KeyError(
+                f"Data dimension field {self.name} is not registered. "
+                f"Call DataDimensionsField.register({self.name})."
+            )
```
romanc: How about we flip this?

```suggestion
        if self.name not in DataDimensionsField._type_registrar:
            raise KeyError(
                f"Data dimension field {self.name} is not registered. "
                f"Call DataDimensionsField.register({self.name})."
            )
        return DataDimensionsField._type_registrar[self.name]
```

ndsl/quantity/data_dimensions_field.py
```diff
@@ -0,0 +1,257 @@
...
+
+        data_dims_size = []
+        for ddim_name in data_dimensions_names:
+            try:
+                data_dims_size.append(quantity_factory.sizer.data_dimensions[ddim_name])
+            except KeyError:
+                raise KeyError(
+                    f'Data dimension axis "{ddim_name}" is not present in QuantityFactory. '
+                    "Use QuantityFactory.add_data_dimensions prior to registering field."
+                )
```
romanc: How about we flip this?

```suggestion
            if ddim_name not in quantity_factory.sizer.data_dimensions:
                raise KeyError(
                    f'Data dimension axis "{ddim_name}" is not present in QuantityFactory. '
                    "Use QuantityFactory.add_data_dimensions prior to registering field."
                )
            data_dims_size.append(quantity_factory.sizer.data_dimensions[ddim_name])
```

romanc (commented):
Looking good to me - just a bunch of nitpicks inline.

ndsl/internal/deferred_type.py
FlorianDeconinck: This is for DaCe, push container/array dimensions resolutions for JIT

ndsl/quantity/data_dimensions_field.py
FlorianDeconinck: I don't have this reflex but yes, indeed!

ndsl/quantity/data_dimensions_field.py
FlorianDeconinck: I should, though this will probably always be used in a loop bound (until it is not for some odd reasons!).

ndsl/quantity/data_dimensions_field.py
```diff
@@ -0,0 +1,257 @@
...
+    def mapping(self) -> SparseNameMapping:
+        return self._mapping
+
+    @mapping.setter
+    def mapping(self, mapping: SparseNameMapping) -> None:
+        self._mapping = mapping
+
+    def index(self, name: str) -> int:
+        if len(self.data_dims) != 1:
+            raise NotImplementedError("Unimplemtened index on a multiple data dims")
```
twicki, line 55:
```suggestion
            raise NotImplementedError("Unimplemented index on a multiple data dims")
```

ndsl/quantity/data_dimensions_field.py
```diff
@@ -0,0 +1,257 @@
...
+        optionally a sparse mapping of name/index.
+
+        The same type cannot be registered twice and will error out.
+
+        Args:
+            pre_registration_type: Type returned by the "declare" function.
+            quantity_factory: Factory carrying the proper data dimensions axis described
+                in `data_dimensions_names`.
+            data_dimensions_names: list of name of data dimension axis.
+            name_mapping: for each dimensions, a sparse dictionnary giving a name/index
```
twicki, line 108:
```suggestion
            name_mapping: for each dimensions, a sparse dictionary giving a name/index
```

tests/quantity/test_datadimensionfields.py
```diff
@@ -0,0 +1,183 @@
...
+def setup_data_dimensions(quantity_factory: QuantityFactory):
+    quantity_factory.add_data_dimensions({"tracers": 8, "plumes": 3})
+
+    # Make sure this is called once
+    global SETUP_DDIMS_ONCE
+    if SETUP_DDIMS_ONCE:
+        return
+    SETUP_DDIMS_ONCE = True
+
+    mappings = {"A": 0, "C": 2, "D": 3, "G": 6, "H": 7}
```
twicki, line 51: I think there is something I don't get: I thought the mapping maps down to a 3d-field?

So for tracer and plumes, is this not illegal?

tests/quantity/test_datadimensionfields.py
```diff
@@ -0,0 +1,183 @@
...
+        self._the_stencil_3D(
+            in_tracers.data[:, :, :, my_index], out_field, self._my_local
+        )
+
+        # Name based access
+        self._the_stencil_3D(
+            in_tracers.data[:, :, :, Tracers.index("H")], out_field, self._my_local
+        )
+
+        # another_index = Tracers.index("H")  # BAD in orchestration
```
twicki, line 105: should this be here as a reminder?

ndsl/quantity/data_dimensions_field.py
```diff
@@ -0,0 +1,257 @@
...
+
+    Methods:
+        register: Register a type by sizing its data dimensions
+        declare: declare a type for future registration
+    """
+
+    _type_registrar: dict[str, _DataDimensionsFieldDescriptor] = {}
+
+    @classmethod
+    def register(
```
twicki, line 90: generally: is there ever a time where we add data dimensions to a quantity-factory but do not want the registered in the Data-Dim-Field? 

If not, should we bundle these up? would require more args to adding ddims. but might make it less error-prone?

FlorianDeconinck:
Requires https://github.com/NOAA-GFDL/NDSL/pull/431 merged first

tests/quantity/test_datadimensionfields.py
FlorianDeconinck: It's the other way around now that you point it out. The restriction in `index` for multiple data dims is wrong. `index` is not dimension-aware, user is in charge to apply it to the right dimensions.


So you can have `{"cloud": 7, "mid" 1}` and do `tracers_and_plumes_qty[:, :, :, Type.index("cloud"), Type.index("mid")]`

ndsl/quantity/data_dimensions_field.py
FlorianDeconinck: It's a good point. I am a bit worried about tying those two API closer. In my mind, this system is far from perfect and probably needs something on top to string it together.

@romanc: opinion?

tests/quantity/test_datadimensionfields.py
FlorianDeconinck: We have expected fail test now

ndsl/quantity/data_dimensions_field.py
FlorianDeconinck: Aaaaaaaaaaand I have a use case: TranslateTest uses one QuantityFactory per rank in serial mode :sob:

romanc: I think there is the _potential_ to streamline workflows here. I write _potential_ because until we have a working version and actual usage in user code, there's only speculation on how exactly `DataDimensionField`s will (have to) be used. While verbose (and thus potentially error prone), I think this is good enough for a first version. Time will tell if we can simplify workflows and/or have use-cases that we currently can't anticipate.

FlorianDeconinck: I would also note that this system is very much the result of the gt4py design of having _static_ data dimension sizes. If we were to curb this, lots of changes (clean up) could flow upstream.

ndsl/stencils/testing/translate.py
```diff
@@ -7,12 +7,15 @@
...
 from ndsl.dsl.stencil import StencilFactory
-from ndsl.optional_imports import cupy as cp
+from ndsl.optional_imports import cupy
 from ndsl.quantity import Quantity
 from ndsl.stencils.testing.grid import Grid
 from ndsl.stencils.testing.savepoint import DataLoader
 
 
+if cupy is None:
+    import numpy as cupy
```
oelbert, line 17: I know why you're doing this but it gives me the heebie-jeebies

ndsl/quantity/data_dimensions_field.py
```diff
@@ -0,0 +1,267 @@
...
+                )
+            data_dims_size.append(quantity_factory.sizer.data_dimensions[ddim_name])
+
+        cls._type_registrar[name] = _DataDimensionDescriptor[
+            gtscript.IJK, (dtype, tuple(data_dims_size))
+        ]
+        if name_mapping is not None:
+            cls._type_registrar[name].mapping = name_mapping
+
+        # Dynamic op replacement fo Type.index() function
```
oelbert: typo

oelbert (changes requested):
one typo

ndsl/stencils/testing/translate.py
FlorianDeconinck: It's not my favorite thing. But until we ramp up `xumpy` as a true cross-device numpy, it must be..

oelbert (approved):

PR #419, opened 2026-04-03
Translate test dimension fix during comparison

CharlesKrop:
added np.squeeze for all non-scalar fields at the top of the BaseMetric class, fixes an error where some fields have a lingering size one dimension causes a dimension mismatch with the reference data. Fields which already have the correct dimensions are unaffected by this change.


FlorianDeconinck (approved):
LGTM - should be covered by tests

PR #431, opened 2026-04-15
[Feature] Implement direct getter/setter for `Quantity`

FlorianDeconinck:
# Description

Dear all, 

We have reached this moment. Yes, after many hours of debate: I have implemented `quantity[:, :, :] = 0` which will behave like `quantity.data[:, :, :] = 0`.

The main reason for the change is orchestration. TL;DR: when parsing, DaCe maps symbols to their name. If a buffer is used with `buffer` and `buffer.data` to access the same memory, it can confuse the parser and lead to difficult to debug numerical error.

This change is also motivated by the fact we have very little to no use for `quantity.field` during runtime (when we might orchestrate).

Further improvement of the API would require the use of `.data` and `.field` to fail during orchestration OR in an "orchestrate" code path. Not implemented here.

## How has this been tested?

The quantity unit test has been changed to not use `.data` and one test has been added.

## Checklist


fmalatino (approved):

tests/quantity/test_quantity.py
```diff
@@ -298,19 +296,35 @@ def test_to_data_array(quantity):
...
+        np.arange(
+            5,
+        ),
+        dims=["dim1"],
+        units="",
+        backend=Backend.python(),
+    )
+
+    assert quantity.shape == quantity.shape
+    assert quantity[3] == quantity[3]
```
romanc: what are these two for?

romanc (approved):
Looks good to me - just a question about some test asserts.

twicki (changes requested):
If we have direct access without `.field` should we at least deprecate it here so we can remove it in the release after that? 

And for my own edification: I thought we specifically put in the `.field` and `.data` functions because that was the only way we could tell orchestration to overwrite the calls for the accesses to work. Why / How did that change? Or do I misremember something here?

tests/quantity/test_quantity.py
FlorianDeconinck: Rename gone wrong! I did the test then `ctrl+d` all `quantity.data` like an animal...

Anyway, there's checking that the accessor changed data properly

FlorianDeconinck:
> If we have direct access without `.field` should we at least deprecate it here so we can remove it in the release after that?

This doesn't really deprecate those access, what it does is make orchestration possible because those access break orchestration. Now - as I said in the PR - what is missing here is a way to forbid those access from being used in orchestrated code path. I tried a few things, it's non-trivial. This alone could warrant a deprecation and removal of the specific accessors.

The counterpoint being that `.field` is mostly what you want when looking in the data outside of the runtime, making it a very useful shortcut.

> And for my own edification: I thought we specifically put in the `.field` and `.data` functions because that was the only way we could tell orchestration to overwrite the calls for the accesses to work. Why / How did that change? Or do I misremember something here?

No, we failed to do so. We couldn't make `dace.views` that worked and we kept having symbol conflicts. Those results are directly motivating this change.

twicki:
sorry, I mistyped in my comment. I ment deprecating `.data` as this use-case is now covered by `[]` access

FlorianDeconinck:
> sorry, I mistyped in my comment. I ment deprecating `.data` as this use-case is now covered by `[]` access

Yes - that we should. We do need an access, so we'll probably (re)introduce a `get_data()` or `get_buffer()` ?

ndsl/quantity/quantity.py
```diff
@@ -329,6 +329,14 @@ def data_as_xarray(self) -> xr.DataArray:
     def np(self) -> ModuleType:
         return self.metadata.np
 
+    def __getitem__(self, subscript: Any) -> Any:
+        """Slicing operator accessing the full buffer"""
+        return self.data[subscript]
+
+    def __setitem__(self, subscript: Any, value: Any) -> None:
+        """Slicing operator setting the full buffer"""
+        self.data[subscript] = value
```
oelbert: should these use `self._data` so we can remove `self.data` more easily in the future?

oelbert (changes requested):
1 minor correction. I do think we should deprecate `quantity.data` but ideally over multiple releases, and as long as we keep `quantity.field`.

FlorianDeconinck:
> 1 minor correction. I do think we should deprecate `quantity.data` but ideally over multiple releases, and as long as we keep `quantity.field`.

The problem with keeping `quantity.field` is that _it will not orchestrate_ in the runtime code. So far, I have no way to warn/cancel it when orchestration. Whereas `quantity[halo:-halo-1, halo:-halo-1, -1]`, the `field` use case, will work but is ugly.

Unsolved issue.

romanc:
Regarding deprecating `.field`: Quantities have a reference to the `Backend`, which knows if we want to orchestrate or not. Could we add a `DeprecationWarning`  inside the `.field` acessor of ` Quantity` depending on the backend? I guess that would only trigger upon usage (and not up-front), but better late than never?

oelbert (approved):
I don't know what to say about the `field` issue beyond just telling people "don't use it at runtime!"

twicki (approved):

FlorianDeconinck:
Holding merge to figure out if we can do something clever for `field`. Like send a picture of Oliver with frown eyebrows with every cloned repository

FlorianDeconinck:
Drafting: I need to go clean up all the `quantity.data` inside `ndsl` itself.

FlorianDeconinck:
Further cleanup:
- :broom: removed all use of `.data[...]` in NDSL
- :lady_beetle: fixed reduce comm bug that was swapping buffer instead of copying. Though it is not per se a bug, this would absolutely fail orchestration
- :heavy_plus_sign: added a `dtype` which seems like a fair re-exposure of the underlying memory. Kept `strides` and `itemsize` hidden, reachable via `_data` because those shouldn't be used directly by the user

FlorianDeconinck:
- :heavy_plus_sign: Restored `sel` API. It is niche but quite useful if you want to slice an array by dimension name without knowing it's indexing
- :heavy_plus_sign: Added a `swap_buffer` methods that replaced the `.data` setter
- :mechanical_arm: Made `hashing` of the `State` better by actually using the `__hash__` of the attributes (which means calling the custom hash of `Quantity`)
- :broom: clean up the `.data` in the unit tests

FlorianDeconinck:
@romanc / @twicki : lots of clean up. please re-review

ndsl/quantity/quantity.py
```diff
@@ -275,10 +276,27 @@ def field(self) -> np.ndarray | cupy.ndarray:
...
+            "the quantity, e.g. `my_quantity[:]` instead of `my_quantity.data[:]`",
+            category=UserWarning,
+            stacklevel=2,
+        )
         return self._data
 
     @data.setter
     def data(self, input_data: np.ndarray | cupy.ndarray) -> None:
+        warnings.warn(
+            "Quantity.data settor is now deprecated. Build a quantity from a data with the "
```
romanc:
```suggestion
            "Quantity.data setter is now deprecated. Build a quantity from a data with the "
```

ndsl/quantity/quantity.py
```diff
@@ -275,10 +276,27 @@ def field(self) -> np.ndarray | cupy.ndarray:
...
+            "dedicated constructor. If you need no-copy mapping, talk to the team.",
+            category=UserWarning,
+            stacklevel=2,
+        )
+        self.swap_buffer(input_data)
+
+    def swap_buffer(self, input_data: np.ndarray | cupy.ndarray) -> None:
+        """Swap internal buffer for given input. Use with _extreme_ care as it might
+        trip hash calculations for other subsystem."""
```
romanc:
```suggestion
        trip hash calculations for other subsystems."""
```

romanc (approved):
Thanks for all the cleanup and hardening. Still looks good to me.


Parallel NDSL unit tests suggest that there's residual use of `Quantity.data` in

- `/home/runner/work/NDSL/NDSL/tests/mpi/test_eta.py`
- `/home/runner/work/NDSL/NDSL/tests/mpi/test_mpi_all_reduce_sum.py`

PR #433, opened 2026-04-16
Deprecate `get_precision()` from `ndsl.dsl.typing`

romanc:
# Description

The function `get_precision()` in `ndsl.dsl.typing` is just returning `NDSL_GLOBAL_PRECISION`, defined in `ndsl.dsl`. Let's use `NDSL_GLOBAL_PRECISION` directly and deprecate the function.

In addition, this PR removes the `TypeAlias` type hints of `global_set_precision()` because they are wrong. For `mypy`, a `TypeAlias` has to be statically known. In our case, `Float` and `Int` depend on `NDSL_GLOBAL_PRECISION`, which is runtime dependent, see [here](https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases) for some examples. Even if there are only two possibilities (64 and 32 bit), `mypy` is stubborn about this and won't even allow a union type. The PR thus suggests to drop the type hint, which allows IDEs and/or language servers to infer the type if they wish.

Changes in `ndsl/grid/geometry.py` and `ndsl/xumpy/minmax.py` are unrelated to the above changes, but also strive to improve type hints.

## How has this been tested?

All good as long as CI is still green.

## Checklist

### DIFF
--- a/ndsl/xumpy/minmax.py
+++ b/ndsl/xumpy/minmax.py
@@ -1,10 +1,10 @@
 from functools import singledispatch
+from typing import Any
 
 import numpy as np
 import numpy.typing as npt
 
 from ndsl import Quantity
-from ndsl.dsl.typing import Float
 from ndsl.optional_imports import cupy as cp
 
 
@@ -58,9 +58,9 @@ def _(in_quantity: Quantity, out_quantity: Quantity) -> None:
 
 @singledispatch
 def max(
-    buffer: npt.NDArray | Quantity | int,
+    buffer: npt.NDArray | Quantity,
     axis: int | tuple[int, ...] | None = None,
-) -> Float:
+) -> Any:
     """Find maximum value."""
     raise NotImplementedError(f"Missing implementation for {type(buffer)}")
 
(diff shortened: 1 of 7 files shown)


ndsl/xumpy/minmax.py
```diff
@@ -58,9 +58,9 @@ def _(in_quantity: Quantity, out_quantity: Quantity) -> None:
 
 @singledispatch
 def max(
-    buffer: npt.NDArray | Quantity | int,
```
romanc, line 61: Not sure where the ` | int` was coming from. I don't think that makes sense

ndsl/xumpy/minmax.py
```diff
@@ -58,9 +58,9 @@ def _(in_quantity: Quantity, out_quantity: Quantity) -> None:
...
+    buffer: npt.NDArray | Quantity,
     axis: int | tuple[int, ...] | None = None,
-) -> Float:
```
romanc, line 63: Imo in case the `buffer` is filled with integers, won't we return an `Int` and not a `Float`? In general, I think, we can't know the return type unless we make this generic on the `buffer`s underlying type.

FlorianDeconinck (approved):
Thanks for the 🧹

PR #438, opened 2026-04-17
[Orchestration] Add `NDSL_VERBOSE_ORCHESTRATION` 

FlorianDeconinck:
# Description

The env var will control a somewhat step-by-step save of SDFGs as it goes down the orchestration optimization. The SDFGs are saved in the same folder than the central SDFG cache (`.dacecache` by default)

Also:
 - Rename option for force GPU sync in orchestration

## How has this been tested?

No-op - debug only.

### DIFF
--- a/ndsl/dsl/dace/dace_config.py
+++ b/ndsl/dsl/dace/dace_config.py
@@ -29,12 +29,14 @@ if TYPE_CHECKING:
 DEACTIVATE_DISTRIBUTED_DACE_COMPILE = False
 
 
-def _debug_dace_orchestration() -> bool:
+def _sync_gpu_option() -> bool:
     """
+    Force synchronize after each kernel call.
+
     Debugging Dace orchestration deeper can be done by turning on `syncdebug`.
     We control this Dace configuration below with our own override.
     """
-    return os.getenv("NDSL_DACE_DEBUG", "False") == "True"
+    return os.getenv("NDSL_DACE_FORCE_SYNC_GPU", "False") == "True"
 
 
 def _is_corner(rank: int, partitioner: Partitioner) -> bool:
@@ -203,6 +205,11 @@ class DaceConfig:
         else:
             self._orchestrate = orchestration
 
+        # Verbose orchestration optimization
+        self.verbose_orchestration = (
+            os.getenv("NDSL_VERBOSE_ORCHESTRATION", "False") == "True"
+        )
+
         # We hijack the optimization level of GT4Py because we don't
         # have the configuration at NDSL level, but we do use the GT4Py
         # level
@@ -323,7 +330,7 @@ class DaceConfig:
 
             # Enable to debug GPU failures
             dace.config.Config.set(
-                "compiler", "cuda", "syncdebug", value=_debug_dace_orchestration()
+                "compiler", "cuda", "syncdebug", value=_sync_gpu_option()
             )
 
             if NDSL_GLOBAL_PRECISION == 32:
--- a/ndsl/dsl/dace/dace_config.py
+++ b/ndsl/dsl/dace/dace_config.py
@@ -205,9 +205,12 @@ class DaceConfig:
         else:
             self._orchestrate = orchestration
 
-        # Verbose orchestration optimization
+        # Verbose optimizations
         self.verbose_orchestration = (
             os.getenv("NDSL_VERBOSE_ORCHESTRATION", "False") == "True"
+        )
+        self.verbose_schedule_tree_optimizations = (
+            os.getenv("NDSL_VERBOSE_SCHEDULE_TREE_OPTIMIZATIONS", "False") == "True"
         )
 
         # We hijack the optimization level of GT4Py because we don't
--- a/ndsl/dsl/dace/orchestration.py
+++ b/ndsl/dsl/dace/orchestration.py
@@ -178,10 +178,12 @@ def _build_sdfg(
                 # Break all loops into uni-dimensional loops to simplify optimizations
                 sdfg.apply_transformations_repeated(MapExpansion, validate=True)
                 stree = sdfg.as_schedule_tree()
-                with open(
-                    os.path.abspath(f"{sdfg.build_folder}/03-pre_opt.stree.txt"), "w"
-                ) as f:
-                    f.write(stree.as_string())
+                if config.verbose_orchestration:
+                    with open(
+                        os.path.abspath(f"{sdfg.build_folder}/02-pre_opt.stree.txt"),
+                        "w+",
+                    ) as f:
+                        f.write(stree.as_string())
 
             with DaCeProgress(config, "Schedule Tree: optimization"):
                 passes = []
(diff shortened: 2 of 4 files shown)


romanc (approved):
Looks good. Should we reuse the same flag to trigger writing debug strees

https://github.com/NOAA-GFDL/NDSL/blob/6110c2eeaa14164e4867da64b1aaceca99a58908/ndsl/dsl/dace/stree/pipeline.py#L51-L53

in the stree opt pipeline?

FlorianDeconinck:
Added `stree` save and also dumping the static memory report in the .dacecache, because why not

ndsl/dsl/dace/orchestration.py
```diff
@@ -158,15 +158,30 @@ def _build_sdfg(
...
+                )
 
         if _INTERNAL__SCHEDULE_TREE_OPTIMIZATION:
             # Here be 🐉 - but tests exists in test_optimization.py
             with DaCeProgress(config, "Schedule Tree: generate from SDFG"):
                 # Break all loops into uni-dimensional loops to simplify optimizations
                 sdfg.apply_transformations_repeated(MapExpansion, validate=True)
                 stree = sdfg.as_schedule_tree()
+                with open(
+                    os.path.abspath(f"{sdfg.build_folder}/03-pre_opt.stree.txt"), "w"
```
romanc: What is number 02?

Also, use "w+" mode to create the file (it won't always already exist). This is what the failing test complain about.

ndsl/dsl/dace/dace_config.py
```diff
@@ -29,12 +29,14 @@
...
-def _debug_dace_orchestration() -> bool:
+def _sync_gpu_option() -> bool:
     """
+    Force synchronize after each kernel call.
+
     Debugging Dace orchestration deeper can be done by turning on `syncdebug`.
     We control this Dace configuration below with our own override.
     """
-    return os.getenv("NDSL_DACE_DEBUG", "False") == "True"
+    return os.getenv("NDSL_DACE_FORCE_SYNC_GPU", "False") == "True"
```
romanc, line 39: I agree that formal deprecation would be overkill for the rename of this debug flag. Let's just make sure that we remember this breaking change for the release notes.

romanc (commented):
Adding the schedule trees from the orchestration is a nice idea. I meant to rezse that env var to control the verbose printing of strees after wvery stree optimization pass (which is currently on by default if stree opt is turned on, e.g. in tests). If we are printing the stree before/after the opt pipeline, I'd suggest to not print the strees per opt pipeline step anymore (unless explicitly activated).

FlorianDeconinck:
Verbosed the `stree` pass into a `NDSL_VERBOSE_SCHEDULE_TREE_OPTIMIZATIONS` because I was there and your comment make sense.

When we clean up our optimization, it'll be time to turn all of this to a level of verbosity

ndsl/dsl/dace/stree/pipeline.py
```diff
@@ -26,8 +26,11 @@ def run(
 
 class CPUPipeline(StreePipeline):
     def __init__(
-        self, passes: list[stree.ScheduleNodeTransformer] | None = None
+        self,
+        passes: list[stree.ScheduleNodeTransformer] | None = None,
+        cache_directory: str = "./",
```
romanc: how about using `pathlib.Path` objects instead of strings?


```suggestion
        cache_directory: Path | None = None,
```

ndsl/dsl/dace/stree/pipeline.py
```diff
@@ -26,8 +26,11 @@ def run(
...
     ) -> None:
+        self.cache_directory = cache_directory
         self.passes = (
             passes if passes is not None else [CartesianAxisMerge(AxisIterator._K)]
         )
```
romanc: It looks like we are starting to duplicate things in the constructors of the CPU- & GPU-Pipeline. How about moving `self.cache_directory`  and ` self.passes` up to the abstract base class?

After doing so, looking at the current implementation, it looks like we can generalize the `__repr__` and `__hash__`  functions. Even the ` run()`  method could be lifted to the base class. The whole file could look something like this, I think

```py
from abc import ABC
from pathlib import Path

import dace.sdfg.analysis.schedule_tree.treenodes as stree

from ndsl.dsl.dace.stree.optimizations import AxisIterator, CartesianAxisMerge
from ndsl.logging import ndsl_log_on_rank_0


class StreePipeline(ABC):
    def __init__(
            self,
            *,
            passes: list[stree.ScheduleNodeTransformer],
            cache_directory: Path | None = None,
    ) -> None:
        if cache_directory is None:
            cache_directory = Path()

        self.cache_directory = cache_directory
        self.passes = passes

    def __hash__(self) -> int:
        return hash(repr(self))

    def __repr__(self) -> str:
        return str([type(p) for p in self.passes])

    def run(
        self,
        stree: stree.ScheduleTreeRoot,
        verbose: bool = False,
    ) -> stree.ScheduleTreeRoot:
        for i, p in enumerate(self.passes):
            if verbose:
                path = self.cache_directory / "pass{i}_{p}.txt"
                ndsl_log_on_rank_0.info(f"[Stree OPT] {p} (saving {path} after)")

            p.visit(stree)

            if verbose:
                with open(path, "w+") as f:
                    f.write(stree.as_string())

        return stree


class CPUPipeline(StreePipeline):
    def __init__(
        self,
        *,
        passes: list[stree.ScheduleNodeTransformer] | None = None,
        cache_directory: Path | None = None,
    ) -> None:
        super().__init__(
            passes=passes if passes is not None else [CartesianAxisMerge(AxisIterator._K)],
            cache_directory=cache_directory,
        )

class GPUPipeline(StreePipeline):
    def __init__(
        self,
        passes: list[stree.ScheduleNodeTransformer] | None = None,
        cache_directory: Path | None = None,
    ) -> None:
        super().__init__(
            passes=passes if passes is not None else [],
            cache_directory=cache_directory,
        )

```

ndsl/dsl/dace/orchestration.py
```diff
@@ -241,7 +271,11 @@ def _build_sdfg(
 
         with DaCeProgress(config, "Simplify (2)"):
             _simplify(sdfg)
-
+            if config.verbose_orchestration:
+                sdfg.save(
+                    os.path.abspath(f"{sdfg.build_folder}/06-simplify_2.sdfgz"),
```
romanc:
```suggestion
                    os.path.abspath(f"{sdfg.build_folder}/05-simplify_2.sdfgz"),
```

romanc (commented):
More ideas for cleanup ...

ndsl/dsl/dace/stree/pipeline.py
FlorianDeconinck: In this implementation (which I have pushed) StreePipeline is no longer an ABC (which is fine). Arguably we could reduce further and take out the Specialization by passing the Backend and altering the "good default"

Time will tell on this

PR #451, opened 2026-05-04
cleanup: type fixes in `Communicator` classes

romanc:
# Description

This PR brings typing fixes for the `Communicator` classes. Python static analysis tools (e.g. `pyright` and `mypy`) are very strict about overloading class members in subclasses, see https://github.com/microsoft/pyright/issues/7840.

This PR suggests to the `Communicator` base class a `Generic` on the partitioner. That allows us to tell static analysis tools that `TileCommunicators` come with a `TilePartitioner` and `CubedSphereCommunicators` come with a `CubedSpherePartitioner`.

Minor other type fixes are based on the fact that overload function signatures should have arguments with the same name as the base class, e.g. `subtile_extent()` of different `Partitioners`.

## How has this been tested?

All good as long as CI is still happy.

## Checklist

### DIFF
--- a/ndsl/comm/communicator.py
+++ b/ndsl/comm/communicator.py
@@ -78,12 +78,12 @@ class Communicator(abc.ABC, Generic[P]):
 
     @property
     def rank(self) -> int:
-        """rank of the current process within this communicator"""
+        """Rank of the current process within this communicator."""
         return self.comm.Get_rank()
 
     @property
     def size(self) -> int:
-        """Total number of ranks in this communicator"""
+        """Total number of ranks in this communicator."""
         return self.comm.Get_size()
 
     def _maybe_force_cpu(self, module: ModuleType) -> ModuleType:
@@ -104,7 +104,7 @@ class Communicator(abc.ABC, Generic[P]):
     def _create_all_reduce_quantity(
         self, input_metadata: QuantityMetadata, input_data: Any
     ) -> Quantity:
-        """Create a Quantity for all_reduce data and metadata"""
+        """Create a Quantity for all_reduce data and metadata."""
         all_reduce_quantity = Quantity(
             input_data,
             dims=input_metadata.dims,
@@ -649,7 +649,7 @@ def bcast_metadata(comm: CommABC, array: Quantity):  # type: ignore[no-untyped-d
 
 
 class TileCommunicator(Communicator[TilePartitioner]):
-    """Performs communications within a single tile or region of a tile"""
+    """Performs communications within a single tile or region of a tile."""
 
     def __init__(
         self,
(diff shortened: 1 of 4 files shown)


ndsl/comm/communicator.py
```diff
@@ -663,10 +666,7 @@ def __init__(
             force_cpu: force all communication to go through central memory
             timer: Time communication operations.
         """
-        super(TileCommunicator, self).__init__(
```
romanc, line 666: listing `(TileCommunicator, self)` as arguments of `super()` is not necessary (anymore).

ndsl/comm/communicator.py
```diff
@@ -794,12 +792,9 @@ def __init__(
                 f"comm object with only {comm.Get_size()} ranks, are we running "
                 "with mpi and the correct number of ranks?"
             )
-        self._tile_communicator: TileCommunicator | None = None
-        self._force_cpu = force_cpu
```
romanc, line 798: unused duplicate member variable

ndsl/comm/communicator.py
```diff
@@ -663,10 +666,7 @@ def __init__(
             force_cpu: force all communication to go through central memory
             timer: Time communication operations.
         """
-        super(TileCommunicator, self).__init__(
-            comm, partitioner, force_cpu=force_cpu, timer=timer
-        )
-        self.partitioner: TilePartitioner = partitioner
```
romanc, line 669: Now unneeded because the correct type is on the base class already.

ndsl/comm/communicator.py
```diff
@@ -814,55 +809,53 @@ def from_layout(
...
-    def _initialize_tile_communicator(self) -> None:
-        tile_comm = self.comm.Split(
-            color=self.partitioner.tile_index(self.rank), key=self.rank
-        )
-        self._tile_communicator = TileCommunicator(tile_comm, self.partitioner.tile)
+        return self._tile_communicator
 
     def _get_gather_recv_quantity(
-        self, global_extent: Sequence[int], metadata: QuantityMetadata
+        self, global_extent: Sequence[int], send_metadata: QuantityMetadata
```
romanc, line 822: Same argument name as base class method.

ndsl/comm/partitioner.py
```diff
@@ -160,9 +156,7 @@ def subtile_index(self, rank: int) -> tuple[int, int]:
     def total_ranks(self) -> int:
         return self.layout[0] * self.layout[1]
 
-    def global_extent(
-        self, rank_metadata: Quantity | QuantityMetadata
-    ) -> tuple[int, ...]:
```
romanc, line 165: Base class only works with `QuantityMetadata`. This technically also works with the `Quantity` because we forward properties like `.dims` and `.extent` to the `Quantity`, but I guess consistency is key here.

We can have a separate discussion about why we have `QuantityMetadata` when one can basically access every member of the meta data directly on the `Quantity`.

romanc (commented):
some remarks inline

FlorianDeconinck (approved):
🧹 🧹 Thanks ! 🧹🧹

PR #452, opened 2026-05-04
[Fix | Memory] NDSL pass down normalized dimension names to `gt4py.cartesian`

FlorianDeconinck:
# Description

`gt4py.cartesian` expects for its dimension names in allocator:
- "I", "J" or "K" - e.g. the "cartesian" dimensions
- size-as-string for all other dimensions

Previous code was failing in the case where a non-cartesian dimension was given first because of the complicated code trying to tap directly into the cartesian naming. We re-write this normalization step to enforce the above without expecting a particular order in the given dimensions.

## How has this been tested?

CI

## Checklist

### DIFF
--- a/ndsl/quantity/quantity.py
+++ b/ndsl/quantity/quantity.py
@@ -91,18 +91,20 @@ class Quantity:
         is_optimal_layout = gt4py_backend_cls.storage_info["is_optimal_layout"]
         device = gt4py_backend_cls.storage_info["device"]
 
-        dimensions: tuple[str | int, ...] = tuple(
-            [
-                (
-                    axis  # type: ignore # mypy can't parse this list construction of hell
-                    if any(dim in axis_dims for axis_dims in constants.SPATIAL_DIMS)
-                    else str(data.shape[index])
-                )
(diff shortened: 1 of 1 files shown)


FlorianDeconinck:
@twicki Romand and I drafted that code - give us another pair of :eye: 

Relevant code in gt4py -> https://github.com/GridTools/gt4py/blob/c7d162cccc35cb2d1aaa79f5ad12222f617803ac/src/gt4py/storage/cartesian/utils.py#L113

romanc (approved):
👍

twicki (commented):
If it works that is great. But I don't understand how that shuffles the data dim? This really only removes `_interface` but leaves the oder the same, right?

FlorianDeconinck:
> If it works that is great. But I don't understand how that shuffles the data dim? This really only removes `_interface` but leaves the oder the same, right?

The bug was not on the order, the bug was that because we _specifically_ looked for cartesian dimensions first, this started not working as soon as we had non-aligned memory - which the work on layout created. It was always a bug, we got lucky.

With this version, you are allowed to pass dimensions in _any_ order.

Issue #440, opened 2026-04-20
Failure to load DaCe does not trigger reasonable NDSL error


CharlesKrop:
If NDSL runs into a problem while loading the DaCe dependency, it does not crash with a reasonable error - things continue moving along and break in a misleading way later on.

For me, this occurred with a missing `yaml` dependency, which caused `import dace` to break (with a reasonable error stating that I needed `yaml`), but importing NDSL or `from ndsl import Backend` produced the following message: `ValueError: NDSL backend orch:dace:cpu:IJK does not have a working GT4Py version. GT4Py backend dace:cpu_kfirst is not registered. Contact the team.`

It is unclear if this also happens with other dependencies, but I argue that NDSL should have a mechanic to explain which dependencies are missing/failing, or stop and report the error from below/within the failing dependency.

romanc:
closing as done via PR #453.

PR #453, opened 2026-05-05
QOL: detect missing dace backends in GT4Py

romanc:
# Description

According to issue https://github.com/NOAA-GFDL/NDSL/issues/440, a missing DaCe installation is detected rather late and the resulting error from the backend initialization is confusing. This PR suggests to fix that by explicitly checking in `ndsl/dsl/__init__.py` if GT4Py could load the DaCe backend. That's the same point where we also set the literal precision.

If we detect missing DaCe backends in GT4Py, a descriptive exception is raised:

```shell
(.venv) romanc@localhost /tmp/NDSL (romanc/detect-dace-missing) $ python
Python 3.13.12 (main, Feb 09 2026, 22:37:44) [GCC] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import ndsl
/tmp/NDSL/.venv/lib64/python3.13/site-packages/gt4py/cartesian/__init__.py:13: UserWarning: GT4Py was unable to load DaCe. DaCe backends (`dace:cpu`, `dace:cpu_kfirst`, and `dace:gpu`) will not be available.
  from . import (
Traceback (most recent call last):
  File "<python-input-0>", line 1, in <module>
    import ndsl
  File "/tmp/NDSL/ndsl/__init__.py", line 2, in <module>
    from . import dsl  # isort:skip
    ^^^^^^^^^^^^^^^^^
  File "/tmp/NDSL/ndsl/dsl/__init__.py", line 56, in <module>
    raise RuntimeError(
        "NDSL installation is incomplete: GT4Py was unable to load the DaCe backends."
    )
RuntimeError: NDSL installation is incomplete: GT4Py was unable to load the DaCe backends.
```

The typo fix in `ndsl/config/backend.py` is unrelated.

@CharlesKrop / @FlorianDeconinck what do you think? Good enough (for now)?

## How has this been tested?

I've checked out NDSL into a separate folder, initialized the submodules, ran `pip install .` and then `pip uninstall dace` to get to a working state without the DaCe backends loaded in GT4Py.

## Checklist

### DIFF
--- a/ndsl/config/backend.py
+++ b/ndsl/config/backend.py
@@ -76,7 +76,7 @@ class Backend:
     on the frontend code. Additionally, it gives a hint toward the macro-strategy
     for loop ordering (IJK, KJI, etc.) or a more broad intent (debug, numpy).
 
-    For convenience, shorcuts are given to the most common needs (
+    For convenience, shortcuts are given to the most common needs (
     `backend_python`, `backend_cpu`, `backend_gpu`).
     """
 
--- a/ndsl/dsl/__init__.py
+++ b/ndsl/dsl/__init__.py
@@ -1,6 +1,7 @@
 # Literal precision for both GT4Py & NDSL
 import os
 import sys
+import warnings
 from typing import Literal
 
 from ndsl.comm.mpi import MPI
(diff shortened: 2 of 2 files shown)


FlorianDeconinck (approved):
Looking good.

Side note: I wonder if we should remove the `traceback` in "production" mode, if we had a prod mode. People tend to be very bad at reading traceback and somewhat worse when it's python's because they are wordy instead of file + line.

romanc:
> Side note: I wonder if we should remove the `traceback` in "production" mode, if we had a prod mode. People tend to be very bad at reading traceback and somewhat worse when it's python's because they are wordy instead of file + line.

Imo this is worth keeping in mind for once we have a production mode.

PR #454, opened 2026-05-05
New Basic Operations

CharlesKrop:
Added 2D and 3D addition, subtraction, multiplication, division and a 2D copy (3D already exists) to `basic_operations.py`. 

There are two sets of the mathematics stencils, one which modify inputs (`a = a + b`) and one which do not (`a + b = c`).

New stencils are covered by tests in `test_basic_operations.py`.


tests/test_basic_operations.py
```diff
@@ -3,24 +3,133 @@
...
+    multiply_2d,
     set_value,
+    subtract,
+    subtract_2d,
 )
 
 
 class Copy:
     def __init__(self, stencil_factory: StencilFactory):
         grid_indexing = stencil_factory.grid_indexing
```
romanc: Any specific reason to keep `grid_indexing` around? As far as I can see it's unused now. Also applies to other test classes.

tests/test_basic_operations.py
```diff
@@ -3,24 +3,133 @@
...
+    divide_2d,
+    multiply,
+    multiply_2d,
     set_value,
+    subtract,
+    subtract_2d,
 )
 
 
 class Copy:
```
romanc: Not your code, but we could make all of those derive from `NDSLRuntime`, i.e.


```suggestion
class Copy(NDSLRuntime):
```

romanc (approved):
Looks good to me. Thanks for adding those.

We could talk about consistency, e.g.

- `add_2d` vs `set_value_2D`
- `adjustmentfactor_stencil` vs `adjust_divide_stencil` - imo we could even loose the `_stencil` suffix
-  `set_IJ_mask_value` also looks out of sync

However, those were all previously there and would need - to be nice with our users - properly deprecated for a release before being removed.

ndsl/stencils/basic_operations.py
```diff
@@ -12,78 +12,198 @@
...
+        input: input field
+        output: output field
     """
     with computation(PARALLEL), interval(...):
-        q_out = q_in
+        output = input
 
 
-def adjustmentfactor_stencil(adjustment: FloatFieldIJ, q_out: FloatField) -> None:
+def copy_2d(input: FloatFieldIJ, output: FloatFieldIJ) -> None:
```
twicki: what is the use-case for this?
If you call this stencil with two 2D-Fields you can call it on `interval(...)` so it matches the 3d stencil if that is done forward as well.

ndsl/stencils/basic_operations.py
```diff
@@ -12,78 +12,198 @@
...
+    with computation(FORWARD), interval(0, 1):
+        output = input
+
+
+def add(input_1: FloatField, input_2: FloatField, output: FloatField) -> None:
+    """
+    Add two inputs together, output to a new field.
+
+    Args:
+        input_1: input field
```
twicki:
```suggestion
        summand_1: input field
        summand_2: input field
        sum: input field
```

ndsl/stencils/basic_operations.py
```diff
@@ -12,78 +12,198 @@
...
+        input_2: input field
+        output: output field
     """
     with computation(PARALLEL), interval(...):
-        q_out = q_out * adjustment
+        output = input_1 + input_2
 
 
-def set_value(q_out: FloatField, value: Float) -> None:
+def add_2d(input_1: FloatFieldIJ, input_2: FloatFieldIJ, output: FloatFieldIJ) -> None:
```
twicki: same question as above

ndsl/stencils/basic_operations.py
```diff
@@ -12,78 +12,198 @@
...
+    input_1: FloatFieldIJ, input_2: FloatFieldIJ, output: FloatFieldIJ
+) -> None:
     """
-    Sets every element of buffer to the value specified by value argument.
+    Subtract input_2 from input_1, output to a new field - 2D variant.
+
+    Args:
+        input_1: input field
+        input_2: input field
+        output: output field
```
twicki:
```suggestion
        minuend: input field
        subtrahend: input field
        difference: output field
```

ndsl/stencils/basic_operations.py
```diff
@@ -12,78 +12,198 @@
...
+
+
+def multiply(input_1: FloatField, input_2: FloatField, output: FloatField) -> None:
+    """
+    Multiply two inputs together, output to a new field.
+
+    Args:
+        input_1: input field
+        input_2: input field
+        output: output field
```
twicki:
```suggestion
        factor_1: input field
        factor_2: input field
        product: output field
```

ndsl/stencils/basic_operations.py
```diff
@@ -12,78 +12,198 @@
...
+
+def divide(input_1: FloatField, input_2: FloatField, output: FloatField) -> None:
+    """
+    Divide input_1 by input_2, output to a new field.
 
     Args:
-        buffer: output field
+        input_1: input field
+        input_2: input field
+        output: output field
```
twicki:
```suggestion
        dividend: input field
        divisor: input field
        quotient: output field
```

ndsl/stencils/basic_operations.py
```diff
@@ -12,78 +12,198 @@
...
+    """
+    with computation(FORWARD), interval(0, 1):
+        output = input_1 / input_2
+
+
+def set_value(field: FloatField, value: Float) -> None:
+    """
+    Sets every element of field a value.
```
twicki:
```suggestion
    Sets every element of a field to a single value.
```

ndsl/stencils/basic_operations.py
```diff
@@ -12,78 +12,198 @@
...
+        field: output field
         value: value of Float type
     """
     with computation(FORWARD), interval(0, 1):
-        buffer = value
+        field = value
 
 
-def set_IJ_mask_value(mask_out: BoolFieldIJ, value: Bool) -> None:
+def set_IJ_mask_value(mask: BoolFieldIJ, value: Bool) -> None:
```
twicki: so this is set_value_2D but bool? I would probalby reflect that as compared to a fully new name

tests/test_basic_operations.py
CharlesKrop: Nope, missed it. Gone!

tests/test_basic_operations.py
CharlesKrop: Good idea, will swap them.

ndsl/stencils/basic_operations.py
CharlesKrop: I'm not sure I follow you here. Are you just saying that I should swap `interval(0, 1)` to `interval(...)`? That's fair. it's done.

I don't have a use for the stencil at the moment, but I want to make sure I include 2d variants of everything so people have a full set available.

CharlesKrop:
Back to draft for the weekend - I am adding more stencils to the batch

ndsl/stencils/basic_operations.py
CharlesKrop: reverted interval(...) back to interval(0, 1) to avoid over computing

ndsl/stencils/basic_operations.py
CharlesKrop: cannot change name, this stencil is used in a few places. will make a separate PR at some point handling all of the in-use name changes

ndsl/stencils/basic_operations.py
oelbert: Do we only need this stencil specifically because `input` and `output` are typehinted as 2D?

oelbert: If that's the case why not extend to `IntField` and IntField2d`?

oelbert (commented):
This adds a lot of stencils and I guess I'm curious how much we need to specify each of them here, especially 2D and 3D versions? How often are we only adding or only dividing in our numerics?

CharlesKrop:
> This adds a lot of stencils and I guess I'm curious how much we need to specify each of them here, especially 2D and 3D versions? How often are we only adding or only dividing in our numerics?

Surprisingly often. I found that, when I don't have these stencils, I just end up wrapping the simple one-liner adds into other stencils, even if it doesn't make a lot of sense scientifically for them to be there, but once I had these stencils I started was more free to write discrete scientific blocks full of smaller, more narrowly tuned stencils, which has the added effect of allowing stencils names to more accurately act as (in my opinion) the best documentation.

CharlesKrop:
To you other point - yes and no. 3d vs 2d need to be their own entity, because 3d can use computation/interval `PARALLEL`/`...`, while 2d must use (at least for now) `FORWARD`/`0, 1`.

It would be nice to say "this stencil can use any type/shape so long as all fields are of the same type/shape" and have one stencil work everything, but I think we are quite a way from having something like that.

CharlesKrop:
A good example of their use can be found here: [Fortran](https://github.com/CharlesKrop/GEOSgcm_GridComp/blob/dsl/single_pymoist_call/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSmoist_GridComp/GEOS_MoistGridComp.F90) v [NDSL (WIP)](https://github.com/CharlesKrop/GEOSgcm_GridComp/blob/dsl/single_pymoist_call/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSmoist_GridComp/pyMoist/pyMoist/moist.py)

oelbert (approved):
Once it's up to date I'm happy to merge in

twicki:
We will keep this branch with the breaking changes up for when we need it, but for now we want to merge the nonbreaking part here: https://github.com/NOAA-GFDL/NDSL/pull/499

PR #455, opened 2026-05-05
[Fix | Orchestration] Dynamic type hint removal for quantity

FlorianDeconinck:
# Description

The `Quantity` class has a `__descriptor__` for DaCe to be able to interpret the buffer. But the parser of DaCe expects that if the `__descriptor__` exists on the Type (e.g. while type hinting) it is callable.

Previous attempts at dealing with this was replacing the `Quantity` type hint with a `dace.compiletime`. This is not good enough, as this will lead to symbol duplication depending on the usage of said `Quantity` in code. A better approach, is to _remove_ entirely the type hint - and let the _runtime_ call to `__descriptor__` do its work.

Also in this PR:
- move `__INLINED` keyword into forwarded gtp4y API as this is _still_ useful for compile time code exclusion in stencils

## How has this been tested?

Locally on microphysics

## Checklist

### DIFF
--- a/ndsl/dsl/dace/orchestration.py
+++ b/ndsl/dsl/dace/orchestration.py
@@ -665,7 +665,9 @@ def orchestrate(
 
     # Remove type hint of Quantity to allow for __descriptor__ to be read in JIT
     for arg_name, annotation in func.__annotations__.items():
-        if annotation in [Quantity]:
+        if annotation in [Quantity] or (
+            isinstance(annotation, type) and issubclass(annotation, Quantity)
+        ):
             func.__annotations__[arg_name] = None
 
     # Build DaCe orchestrated wrapper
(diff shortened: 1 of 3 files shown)


twicki (approved):
If this works that is great

romanc (approved):
I've linked this PR from the corresponding issue (https://github.com/NOAA-GFDL/NDSL/pull/455).

I also think we should have a test case for just the `Quantity` case, which now works. So I've added that directly to the PR. The `xfail` for the program with a `State` type still remains (because that's unsolved).

romanc:
What about `Locals`?

FlorianDeconinck:
> What about `Locals`?

Now covered, as well as all derived type of `Quantity`. Covered in utest

PR #450, opened 2026-05-04
ci: run diag manager tests on CI

romanc:
# Description

This is a follow-up from https://github.com/NOAA-GFDL/NDSL/pull/427 attempting to run pyFMS tests on CI. Doing so currently fails with a segfault when attempting to open `diag_manager_single_tile.nc` in line

```py
ds = xr.open_mfdataset("diag_manager_single_tile.nc", decode_times=True)
```

## How has this been tested?

All good as long as CI is green ... 

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (e.g. add new modules to docs/docstrings/)
- [ ] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] New check tests, if applicable, are included

PR #459, opened 2026-05-07
[Fix]: Removing race condition within `DiagManagerMonitor` testing

fmalatino:
# Description

This PR introduces changes that prevent a race condition that exists in `test_dm_monitor_single`. This PR builds off the work introduced by [PR 450](https://github.com/NOAA-GFDL/NDSL/pull/450).

## How has this been tested?

Tested using current CI

## Checklist


.github/workflows/unit_tests.yaml
```diff
@@ -5,8 +5,8 @@ on:
   pull_request: # ... a PR is opened / updated
   merge_group: # ... the PR is added to the merge queue
   push:
-    branches:
-      - main # ... when merging into the main branch
+    # branches:
+    #   - main # ... when merging into the main branch
```
romanc: Why did you comment these two lines?

.github/workflows/unit_tests.yaml
```diff
@@ -31,17 +42,20 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - name: Install mpi (MPICH flavor)
-        run: pip3 install mpich
+      - name: Install dependencies
+        run: |
+          sudo apt-get update --fix-missing
+          sudo apt-get install libopenmpi-dev netcdf-bin libnetcdf-dev libnetcdff-dev nco libyaml-dev diffutils
```
romanc: Two things here:

1. I'd prefer to not do any `apt-get install` in the NDSL pipelines unless we absolutely need them (e.g. in the case of the pyFMS monitor). 
2. I would avoid `apt-get update` in pipelines to keep workflows as reproducible as possible.

```suggestion
      - name: Install MPI
        if: ${{!contains(matrix.extra, 'pyfms')}}
        run: pip3 install mpich

      - name: Install pyFMS dependencies (includes system MPI)
        if: contains(matrix.extra, 'pyfms')
        run: sudo apt-get install libopenmpi-dev netcdf-bin libnetcdf-dev libnetcdff-dev nco libyaml-dev diffutils
```

.github/workflows/unit_tests.yaml
```diff
@@ -31,17 +42,20 @@ jobs:
...
 
       - name: Install Python packages
-        run: pip3 install .[test,zarr]
+        run: |
+          pip3 install -v .[${{matrix.extra}}]
```
romanc: why do you suggest to run this one-line call in script mode?

```suggestion
        run: pip3 install -v .[${{matrix.extra}}]
```

.github/workflows/unit_tests.yaml
```diff
@@ -31,17 +42,20 @@ jobs:
...
 
       - name: Run serial cpu tests
-        run: coverage run --rcfile=pyproject.toml -m pytest -m "not parallel and not gpu" tests
+        run: coverage run --rcfile=pyproject.toml -m pytest -s -m "${{matrix.marker-serial}}" tests
 
       - name: Run parallel cpu tests
-        run: mpiexec -np 6 coverage run --rcfile=pyproject.toml -m mpi4py -m pytest -m "parallel and not gpu" tests
+        run: mpiexec -np 6 --oversubscribe coverage run --rcfile=pyproject.toml -m mpi4py -m pytest -s -m "${{matrix.marker-parallel}}" tests
```
romanc: where does the `--oversubscribe` come from? why did you add `-s`?

tests/monitor/test_dm_monitor_single.py
```diff
@@ -220,7 +219,9 @@ def test_dm_monitor_single_tile():
 
     # check output!
     assert Path("diag_manager_single_tile.nc").exists()
-    ds = xr.open_mfdataset("diag_manager_single_tile.nc", decode_times=True)
+    with dask.config.set(scheduler="synchronous"):
+        ds = xr.open_dataset("diag_manager_single_tile.nc", decode_times=True)
+    print("Dataset opened through xarray")
```
romanc: remove debug print
```suggestion
```

tests/monitor/test_dm_monitor_single.py
```diff
@@ -220,7 +219,9 @@ def test_dm_monitor_single_tile():
 
     # check output!
     assert Path("diag_manager_single_tile.nc").exists()
-    ds = xr.open_mfdataset("diag_manager_single_tile.nc", decode_times=True)
+    with dask.config.set(scheduler="synchronous"):
+        ds = xr.open_dataset("diag_manager_single_tile.nc", decode_times=True)
```
romanc: Are you suggesting that any xarray dataset read in a non-parallel environment would need to be guarded by a `with` statement setting the scheduler to `"synchronous"`? That seems like something to be configured in a more central place to me, no?

romanc (changes requested):
I think we'll need to find a more central place to configure the `dask` scheduler.

.github/workflows/unit_tests.yaml
fmalatino: This was done during testing on within my fork and I forgot to uncomment them. My mistake.

.github/workflows/unit_tests.yaml
fmalatino: I previously had other commands in there and neglected to undo this. Undone now.

.github/workflows/unit_tests.yaml
fmalatino: The `--oversubscribe` flag is for use when openmpi is installed instead of mpich. I have amended the workflow now to use mpich for installs with and without pyfms.

tests/monitor/test_dm_monitor_single.py
fmalatino: When I looked up the errors that were being produced I was pointed in the direction of using this method, albeit I am not sure why this is the case given that it is not in a parallel environment. I think the issue comes down to the call to `decode_times` parameter within `open_dataset`, this is based on the information I have seen thus far on what might be the problem. The most recent commit has removed the use of the dask scheduler and instead will use a call to `MPIComm()._comm.Barrier()`, given that traceback makes references to calls regarding threading.

fmalatino:
Per the issues we are seeing in this PR, I think it is necessary to re-work the installation of pyfms, such that it will use the pip installed netcdf for installation. I will move this PR to draft and note when a PR for updating the install of pyfms is available.

.github/workflows/unit_tests.yaml
```diff
@@ -31,17 +42,29 @@ jobs:
...
 
+      - name: Install pyFMS dependencies (includes system MPI)
+        if: contains(matrix.extra, 'pyfms')
+        run: |
+          sudo apt-get update
+          sudo apt-get install libopenmpi-dev netcdf-bin libnetcdf-dev libnetcdff-dev nco libyaml-dev diffutils
+
       - name: Install Python packages
-        run: pip3 install .[test,zarr]
+        run: pip3 install -v .[${{matrix.extra}}]
```
romanc: I don't think we need to run `pip install` in verbose mode
```suggestion
        run: pip3 install .[${{matrix.extra}}]
```

.github/workflows/unit_tests.yaml
```diff
@@ -31,17 +42,29 @@ jobs:
...
-      - name: Install mpi (MPICH flavor)
+      - name: Install MPI
+        if: ${{!contains(matrix.extra, 'pyfms')}}
         run: pip3 install mpich
 
+      - name: Install pyFMS dependencies (includes system MPI)
+        if: contains(matrix.extra, 'pyfms')
+        run: |
+          sudo apt-get update
+          sudo apt-get install libopenmpi-dev netcdf-bin libnetcdf-dev libnetcdff-dev nco libyaml-dev diffutils
```
romanc, line 53: Why call `apt-get update`? Do you have a need for it? Imo it just makes the workflow even less reproducible because versions of system libraries installed might differ between runs.

```suggestion
        run: sudo apt-get install libopenmpi-dev netcdf-bin libnetcdf-dev libnetcdff-dev nco libyaml-dev diffutils
```

.github/workflows/unit_tests.yaml
romanc: > The `--oversubscribe` flag is for use when openmpi is installed instead of mpich. I have amended the workflow now to use mpich for installs with and without pyfms.

Is this still the case? As far as I can see, we are using MPICH for "normal" runs and OpenMPI for pyFMS runs. And thus need two invocations for running the parallel tests. I don't like that (imo unnecessary) duplication. I don't mind which mpi we use, but I'd like to stick with one if possible.

Also, the `-s` is still in there. Please remove unless there's a good reason
```suggestion
        run: mpiexec -np 6 --oversubscribe coverage run --rcfile=pyproject.toml -m mpi4py -m pytest -m "${{matrix.marker-parallel}}" tests
```

setup.py
```diff
@@ -15,7 +15,7 @@ def local_pkg(name: str, relative_path: str) -> str:
     "cftime",
     "xarray>=2025.01.2",  # datatree + fixes
     "f90nml>=1.1.0",
-    "netcdf4==1.7.2",
+    "netcdf4>=1.7.2",
```
romanc: Is this necessary? @FlorianDeconinck are we ready to give up control over the exact `netcdf4` version?

tests/monitor/test_dm_monitor_single.py
```diff
@@ -220,30 +218,30 @@ def test_dm_monitor_single_tile():
...
-    assert ds["var_3d"].attrs["units"] == "muntin"
-    assert ds["time"].shape == (ntimesteps,)
-    assert ds["time"].dims == ("time",)
-    assert ds["time"].values[0] == cftime.DatetimeNoLeap(2, 1, 1, 2, 1, 1)
-    assert ds["time"].values[1] == cftime.DatetimeNoLeap(2, 1, 1, 3, 1, 1)
-    assert ds["time"].values[2] == cftime.DatetimeNoLeap(2, 1, 1, 4, 1, 1)
-    np.testing.assert_array_equal(ds["var_2d"].values[0, :, :], var2_global.transpose())
-    np.testing.assert_array_equal(ds["var_2d"].values[1, :, :], var2_global.transpose())
-    np.testing.assert_array_equal(ds["var_2d"].values[2, :, :], var2_global.transpose())
+    ds = Dataset("diag_manager_single_tile.nc")
```
romanc, line 221: just out of curiosity: did you figure out what the issue with `xr.open_dataset()` was?

romanc (changes requested):

.github/workflows/unit_tests.yaml
fmalatino: I tried running the tests with just mpich, and the tests would fail, with a message regarding lack of processors for the run. I will remove the `-s` on the incoming commit.

.github/workflows/unit_tests.yaml
fmalatino: I add the verbose flag to track the pyfms install, will remove with the next commit.

.github/workflows/unit_tests.yaml
fmalatino: I will remove this and try the runs without. Previously there were errors in the install and the update seemed to fix this.

tests/monitor/test_dm_monitor_single.py
fmalatino: It invoked some multi-threading methods that were creating some race conditions. I didn't dig into what was specifically happening within the call, and instead looked into other ways to access the data in the files and came across using the netcdf package directly, which seemingly does not use threading.

setup.py
fmalatino: This has been amended.

.github/workflows/unit_tests.yaml
fmalatino: I had to add the `apt-get update` back in as it was necessary to avoid an issue with `libcurl` access.

.github/workflows/unit_tests.yaml
```diff
@@ -31,17 +42,24 @@ jobs:
...
 
       - name: Install Python packages
-        run: pip3 install .[test,zarr]
+        run: pip3 install .[${{matrix.extra}}]
 
       - name: Run serial cpu tests
-        run: coverage run --rcfile=pyproject.toml -m pytest -m "not parallel and not gpu" tests
+        run: coverage run --rcfile=pyproject.toml -m pytest -s -m "${{matrix.marker-serial}}" tests
```
romanc: Please remove the extra `-s` argument 

```suggestion
        run: coverage run --rcfile=pyproject.toml -m pytest -m "${{matrix.marker-serial}}" tests
```

romanc (changes requested):
While there is now a combination of system dependencies and user code that makes the tests pass, I'm still a bit uneasy on this feature. In particular, I'm wondering if there's anything we can learn from this message

Image

which [shows up in pyFMS tests](https://github.com/NOAA-GFDL/NDSL/actions/runs/25697320515/job/75448715648?pr=459) now that they run. Tracing that message back to `FMS`, I get here

https://github.com/NOAA-GFDL/FMS/blob/222f3c6de852b6ba565bd5d6bac55893e92ea106/diag_manager/diag_manager.F90#L4221-L4222

which indicates that something in the setup of the diag_manager isn't as expected. And indeed, in NDSL, `diag_manager.init()` is called without the `time_init` argument

https://github.com/NOAA-GFDL/NDSL/blob/4d43f77a26dd81ab1e189ede6e38b035b456b88b/ndsl/monitor/diag_manager_monitor.py#L35

which `pyFMS` does provide as an option:

https://github.com/NOAA-GFDL/pyFMS/blob/2dd98e645e99bfedd610d057e5aaf9541763ea3d/pyfms/py_diag_manager/diag_manager.py#L77-L80

Do you think the above error message could be related to the crashes we have seen previously? If not, do you think there's a way to configure pyFMS or the pyFMS-monitor to avoid the message (i.e. not use `prepend_date` or supply a `time_init` argument)? It looks like `prepend_date` is an option read from a namelist and it defaults to true

https://github.com/NOAA-GFDL/FMS/blob/222f3c6de852b6ba565bd5d6bac55893e92ea106/diag_manager/diag_manager.F90#L191-L197

.github/workflows/unit_tests.yaml
fmalatino: Removed the flag in the most recent commit.

fmalatino:
> While there is now a combination of system dependencies and user code that makes the tests pass, I'm still a bit uneasy on this feature. In particular, I'm wondering if there's anything we can learn from this message
> 
> Image
> which [shows up in pyFMS tests](https://github.com/NOAA-GFDL/NDSL/actions/runs/25697320515/job/75448715648?pr=459) now that they run. Tracing that message back to `FMS`, I get here
> 
> https://github.com/NOAA-GFDL/FMS/blob/222f3c6de852b6ba565bd5d6bac55893e92ea106/diag_manager/diag_manager.F90#L4221-L4222
> 
> which indicates that something in the setup of the diag_manager isn't as expected. And indeed, in NDSL, `diag_manager.init()` is called without the `time_init` argument
> 
> https://github.com/NOAA-GFDL/NDSL/blob/4d43f77a26dd81ab1e189ede6e38b035b456b88b/ndsl/monitor/diag_manager_monitor.py#L35
> 
> which `pyFMS` does provide as an option:
> 
> https://github.com/NOAA-GFDL/pyFMS/blob/2dd98e645e99bfedd610d057e5aaf9541763ea3d/pyfms/py_diag_manager/diag_manager.py#L77-L80
> 
> Do you think the above error message could be related to the crashes we have seen previously? If not, do you think there's a way to configure pyFMS or the pyFMS-monitor to avoid the message (i.e. not use `prepend_date` or supply a `time_init` argument)? It looks like `prepend_date` is an option read from a namelist and it defaults to true
> 
> https://github.com/NOAA-GFDL/FMS/blob/222f3c6de852b6ba565bd5d6bac55893e92ea106/diag_manager/diag_manager.F90#L191-L197

Looping @rem1776 in here as he brought this feature in and generated this test.

rem1776:
@romanc Yeah skipping the `init_time` argument was intentional. It's more of an optional feature to change the name of your output files so that they include the model start date.

The `init_time` argument used in the `diag_manager_init` routine is only used to prepend the starting date to the output filename. If theres no `init_time` present and `prepend_date` is true, after outputting that note message FMS will set `prepend_date` to false so that it doesn't cause a problem.

We could add a line to set prepend_date to false in the created namelists, which would silence the message but not change any other behavior, if thats preferable. We may want to use this option the future, but I think it would be better to do that via a flag in the monitor's init routine instead of doing it every run.

.github/workflows/unit_tests.yaml
romanc: I see ... I guess we'll need it then.

romanc:
@rem1776 thanks for the background on that `init_time` flag. I was asking because @fmalatino had to modify the test code to change

```patch
- ds = xr.open_mfdataset(filename, decode_times=True)
+ ds = netCDF4.Dataset(filename)
```

and then use `num2date()` in the test code to make them run on CI. Reading via `xarray` would end up in a segfault and traces would point to date/time routines. Since I have no experience with FMS, I found this suspicious and I was looking for anything that would explain why this would happen. Would you expect reading via `xarray` to fail?

> We could add a line to set prepend_date to false in the created namelists, which would silence the message but not change any other behavior, if thats preferable. We may want to use this option the future, but I think it would be better to do that via a flag in the monitor's init routine instead of doing it every run.

I agree that this would be better placed in the monitor's init routine.

fmalatino:
> @rem1776 thanks for the background on that `init_time` flag. I was asking because @fmalatino had to modify the test code to change
> 
> ```diff
> - ds = xr.open_mfdataset(filename, decode_times=True)
> + ds = netCDF4.Dataset(filename)
> ```
> 
> and then use `num2date()` in the test code to make them run on CI. Reading via `xarray` would end up in a segfault and traces would point to date/time routines. Since I have no experience with FMS, I found this suspicious and I was looking for anything that would explain why this would happen. Would you expect reading via `xarray` to fail?
> 
> > We could add a line to set prepend_date to false in the created namelists, which would silence the message but not change any other behavior, if thats preferable. We may want to use this option the future, but I think it would be better to do that via a flag in the monitor's init routine instead of doing it every run.
> 
> I agree that this would be better placed in the monitor's init routine.

Do we want this change in this PR or a subsequent one?

.github/workflows/unit_tests.yaml
```diff
@@ -31,17 +42,26 @@ jobs:
...
+        if: contains(matrix.extra, 'pyfms')
+        run: |
+          sudo apt-get update
+          sudo apt-get install libopenmpi-dev netcdf-bin libnetcdf-dev libnetcdff-dev nco libyaml-dev diffutils
 
       - name: Install Python packages
-        run: pip3 install .[test,zarr]
+        run: |
+          pip3 install .[${{matrix.extra}}]
+          pip3 list
```
romanc: please remove debug `pip list` ...

```suggestion
        run: pip3 install .[${{matrix.extra}}]
```

romanc (approved):
It looks like we are a bit stuck here with this PR. Let me propose the following

1. Follow-up issue for the `init_time` argument
2. Follow-up issue for the crash when using `xarray.open_dataset()`
3. and then we merge this PR as-is since it's anyway an improvement over the current state where pyfms-tests don't even run.

fmalatino:
Issues [467 ](https://github.com/NOAA-GFDL/NDSL/issues/467)and [468](https://github.com/NOAA-GFDL/NDSL/issues/468) have been added to track the `xarray` use issue and default arguments to `DiagManagerMonitor` respectively.

rem1776 (approved):

fmalatino:
Checking if this can be merged?

romanc:
> Checking if this can be merged?

Fine by me and since Ryan also approved, I don't see a reason to wait any longer. Florian won't have time to review.

PR #471, opened 2026-06-05
[Feature | Logging] Clean up multi-rank logging

FlorianDeconinck:
# Description

Using NDSL at scale on many-ranks has a logging issue. Pages and pages of (mostly) the same logs. Going to `NDSL_LOG_LEVEL=Error` kills most of them, but could also swallow the more interesting ones.

We switch here the stance of the logger to _only log on rank 0_ by default. The env var `NDSL_LOG_ALL=True` would revert to the previous wordy log.

In order to keep cleaning logs for production run, we introduce a `NDSL_COMPILER_SILENCE`, which defaults to `False`, that when set to `True` will silence all warnings from the compiler. 

## How has this been tested?

No unit tests so far, just local testing.

## Checklist

### DIFF
--- a/ndsl/dsl/__init__.py
+++ b/ndsl/dsl/__init__.py
@@ -1,4 +1,5 @@
 # Literal precision for both GT4Py & NDSL
+import logging
 import os
 import sys
 from typing import Literal
@@ -59,3 +60,8 @@ if not any([name.startswith("dace") for name in gt_backend.REGISTRY.names]):
 
 
 ndsl_log.info(f"Literal precision: {NDSL_GLOBAL_PRECISION}")
+
+# We remove warnings from the compiler for higher level of logging
+if ndsl_log.level > logging.WARNING:
+    gt4py.cartesian.config.build_settings["extra_compile_args"]["cxx"].append("-w")
+    gt4py.cartesian.config.build_settings["extra_compile_args"]["cuda"].append("-w")
--- a/ndsl/dsl/__init__.py
+++ b/ndsl/dsl/__init__.py
@@ -1,5 +1,4 @@
 # Literal precision for both GT4Py & NDSL
-import logging
 import os
 import sys
 from typing import Literal
@@ -62,6 +61,7 @@ if not any([name.startswith("dace") for name in gt_backend.REGISTRY.names]):
 ndsl_log.info(f"Literal precision: {NDSL_GLOBAL_PRECISION}")
 
 # We remove warnings from the compiler for higher level of logging
-if ndsl_log.level > logging.WARNING:
+NDSL_COMPILER_SILENCE = os.getenv("NDSL_COMPILER_SILENCE", "False").lower() == "true"
+if NDSL_COMPILER_SILENCE:
     gt4py.cartesian.config.build_settings["extra_compile_args"]["cxx"].append("-w")
     gt4py.cartesian.config.build_settings["extra_compile_args"]["cuda"].append("-w")
--- a/ndsl/dsl/dace/dace_config.py
+++ b/ndsl/dsl/dace/dace_config.py
@@ -249,7 +249,7 @@ class DaceConfig:
                 "compiler",
                 "cpu",
                 "args",
-                value=f"-march={march_cpu} -std=c++17 -fPIC {warnings_policy} -Wextra -O{optimization_level} {cxx_defaults.cxx_compile_flags}",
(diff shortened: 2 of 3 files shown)


ndsl/dsl/dace/dace_config.py
```diff
@@ -244,11 +244,12 @@ def __init__(
             march_cpu = "armv8-a" if is_arm_neoverse else "native"
             # Removed --fmath
             cxx_defaults = cxx_compiler_defaults(GT4PY_COMPILE_OPT_LEVEL)
+            warnings_policy = "-Wall" if NDSL_COMPILER_SILENCE else "-w"
```
romanc: Given the changes in ` ndsl/dsl/__init__.py`, shouldn't this be the other way round?


```suggestion
            warnings_policy = "-w" if NDSL_COMPILER_SILENCE else "-Wall"
```

ndsl/dsl/dace/dace_config.py
```diff
@@ -244,11 +244,12 @@ def __init__(
...
             dace.config.Config.set(
                 "compiler",
                 "cpu",
                 "args",
-                value=f"-march={march_cpu} -std=c++17 -fPIC -Wall -Wextra -O{optimization_level} {cxx_defaults.cxx_compile_flags}",
+                value=f"-march={march_cpu} -std=c++17 -fPIC {warnings_policy} -Wextra -O{optimization_level} {cxx_defaults.cxx_compile_flags}",
```
romanc: What about `-Wextra` ?

ndsl/dsl/__init__.py
```diff
@@ -59,3 +59,9 @@ def _get_literal_precision(default: Literal["32", "64"] = "64") -> Literal["32",
 
 
 ndsl_log.info(f"Literal precision: {NDSL_GLOBAL_PRECISION}")
+
+# We remove warnings from the compiler for higher level of logging
+NDSL_COMPILER_SILENCE = os.getenv("NDSL_COMPILER_SILENCE", "False").lower() == "true"
```
romanc, line 64: somewhat unrelated, I'm starting to feel like we should have a simple functions to get environment variables. gt4py.next has those

https://github.com/GridTools/gt4py/blob/d661c1b4d640230468abd21132ab4f1cac3076f5/src/gt4py/next/config.py#L37-L63

ndsl/logging.py
```diff
@@ -20,6 +20,28 @@
...
+            return True
+
+        rank = MPI.COMM_WORLD.Get_rank()
+        if record.levelno >= logging.ERROR:
+            return True
+
+        if rank == 0:
+            return True
+
+        return False
```
romanc:
```suggestion
        return rank == 0
```

romanc (commented):
some thoughts

ndsl/dsl/__init__.py
FlorianDeconinck: We need a yaml config file and to get rid of those env var. I keep piling on because my config branch is not finished, but in truth my vote is for _no_ env var at all as they are difficult to archive for keeping record of simulations

ndsl/dsl/dace/dace_config.py
FlorianDeconinck: Yep yep yep

ndsl/dsl/dace/dace_config.py
FlorianDeconinck: I have mixed feeling about `-Wextra`, especially on generated code. Some of the things like unintialized data are very useful, some like overlapping conditional analysis range from good to noisy and some are downright GNU-weird.

So I'd say no, but I can be convinced.

romanc: I mean it's already/still there at the moment. Given your sentiment and that `-Wall` turns on `-Wunsued` (1), I think we can just get rid of it for now.

```suggestion
                value=f"-march={march_cpu} -std=c++17 -fPIC {warnings_policy} -O{optimization_level} {cxx_defaults.cxx_compile_flags}",
```

(1): https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

romanc (approved):

ndsl/dsl/dace/dace_config.py
FlorianDeconinck: I had missed we had it on :D

Removed, thanks!

PR #485, opened 2026-06-25
add domain_position argument and handle pyFMS domain creation in DiagManagerMonitor

rem1776:
# Description
This PR adds an argument to the `register_axis` function so that it can be passed through to pyFMS, as well as some functions to manage the domain creation and id number.

## How has this been tested?
Updated the existing unit test for the added argument and also tested with my pace branch:
https://github.com/rem1776/pace/tree/dm-pt3

Ran with all conda installed dependencies:
gcc/gfortran 15.2
mpich 5.0.1
python 3.12.13

## Checklist


ndsl/dsl/typing.py
FlorianDeconinck, line 1: This is no longer required - code has changed

ndsl/monitor/diag_manager_monitor.py
```diff
@@ -1,27 +1,134 @@
...
+def set_current_pyfms_domain_id(domain_id: int) -> None:
+    global CURRENT_DOMAIN_ID
+    CURRENT_DOMAIN_ID = domain_id
+
+
+def get_current_pyfms_domain_id() -> int | None:
+    return CURRENT_DOMAIN_ID
+
+
+def initialize_pyfms(
```
FlorianDeconinck, line 33: Fine to be here for now - but I reckon this should end up in a central place if we use `pyFMS` do domain decomposition for runtime halo exchange for example

FlorianDeconinck (changes requested):
Please pull `develop` back in - this should resolve the `typing.py` issue for now

PR #465, opened 2026-05-14
Optimization Cycle I: Loop merge & transient refine

FlorianDeconinck:
# Description

Readying for mainline the following Schedule Tree transform:
- `CartesianMerge`
- `InlineVertical2DWrite`
- `CartesianRefineTransients`

 QOL / Tooling:
 - `LocalOptimization`: allows for an optimization to swap to a different config for a given subset of the code
 - `TreeOptimizationStatistics` will record the before/after count of maps, fors and transients
 - We can now force non-interface dimensions to be padded - iregardless of their alignment with Fortran

:lady_beetle: Regression/ Bugs worked around
- `Local`s are now non-transient in GPU because of bugs showing during tree optimization
- `CartesianRefineTransients` is not applied on GPU - same as above

~:warning:  This PR includes an update to temporary branches of gt4py/dace to consolidate all changes needed for the June presentation~

~- DaCe math functions & symbol replacement: https://github.com/spcl/dace/pull/2364~
~- GT4Py math functions: https://github.com/GridTools/gt4py/pull/2595~

This contains the released GT4Py / DaCe required for the features presented here.

## How has this been tested?

New tests when needed

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (e.g. add new modules to docs/docstrings/)
- [ ] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] New check tests, if applicable, are included

PR #491, opened 2026-06-30
fix: cube-sphere communitcator rank limits

romanc:
# Description

Weaken the cube-sphere communicator hard ranks limit. We need "at least" not "exactly" the rank number. Iirc this has especially been an issue with "io rank" when running integrated into GEOS.

This is part of building back the "opt_cycle_I"-branch, i.e. trying to get https://github.com/NOAA-GFDL/NDSL/pull/465 into a smaller changeset.

## How has this been tested?

Running dynamics as part of the June push. @FlorianDeconinck is there an easy way to test this?

## Checklist

     ???

### DIFF
--- a/ndsl/comm/communicator.py
+++ b/ndsl/comm/communicator.py
@@ -786,7 +786,7 @@ class CubedSphereCommunicator(Communicator[CubedSpherePartitioner]):
                 "Communicator needs to be instantiated with communication subsystem"
                 f" derived from `comm_abc.Comm`, got {type(comm)}."
             )
-        if comm.Get_size() != partitioner.total_ranks:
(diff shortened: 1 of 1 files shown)


FlorianDeconinck:
Not really a good way to test this, apart from a mock MPI and a rather silly test. It is a very "operation" oriented case where you oversubscribe your nodes because you need IO to get nodes, but since we are below the infrastructure layer we can reason on the all structure

FlorianDeconinck (approved):
LGTM

romanc:
Alright - in that case, let's just merge as-is saying that we've tested (extensively) with recent runs integrated in GEOS (which is the final bar anyway).

PR #496, opened 2026-07-01
[Feature | Experimental | Debugger] Automatic debugger on `NDSLRuntime.__call__` and `save_all` feature 

FlorianDeconinck:
# Description

- Automatic patching of `NDSLRuntime.__call__` with the debugger override
- Adding a `save_all` option to the debugger that will dump all inputs/outputs for each stencils
- Add `DebuggerStartFrom` that delay a data save until a certain call count for a named `NDSLRuntime`
- Skip optional argument that are `None` when saving

## How has this been tested?

Unit tests added

## Checklist


ndsl/debug/debugger.py
```diff
@@ -18,14 +18,17 @@ class Debugger:
 
     # Configuration
     stencils_or_class: list[str] = dataclasses.field(default_factory=list)
+    timestep_name: str = ""
```
twicki: As far as I understand this code this is a call-counter not a timestep-counter, right? If so, I would not call it timestep since there is a lot of work that is called multiple times per step.

I also think that a call-counter is the right approach, especially this deep down in debugging we can expect people to know how much their code is called each timestep and do simple arithmetics.

ndsl/dsl/ndsl_runtime.py
```diff
@@ -71,7 +113,10 @@ def check_for_quantity(object_: object) -> None:
             check_for_quantity(self)
 
         # Orchestrate __call__ by default
-        if callable(self):
+        if self._stencil_factory.backend.is_orchestrated() and callable(self):
```
twicki, line 118: is there a reason why we always monkey-patch and only revert for orch instead of checking if not orch before patching?

romanc (approved):
Looks good to me. Thanks for expanding our debug utilities.

ndsl/debug/debugger.py
oelbert: I agree with Tobias, this specific difference gets confused a fair bit and I'd like to be precise with naming here specifically

FlorianDeconinck: It is not a call count, because it's specific to a particular `NDSLRuntime` class name. You could argue timestep == call of the "top level", which is what it is. But this is used to delay debug writing. A more complete approach would allow multiple of those so you can target systems without having to know exactly their architecture, which is useful at scale.

I can side step the naming issue entirely by renaming to something else if the "timestep" wording is to loaded here.

FlorianDeconinck: @oelbert  / @twicki : changes have been made. The new configuration looks like this

```yaml
  stencils_or_class:
    - StencilA
  track_parameter_by_name:
    - alpha
  save_compute_domain_only: True,
  dir_name: my/dir/
  save_all: False,
  save_from: 
    ndslruntime_name: StencilA
    start_from_call: 2
```

ndsl/dsl/ndsl_runtime.py
FlorianDeconinck: Because I don't have the backend when the monkey patching happens. To monkey patch the `.__call__`  you need to do it on the Class, and StencilFactory and/or Backend is not available at this stage

twicki: I understand, thanks

twicki (approved):

Issue #500, opened 2026-07-08
DataDimension size does not support NDSL types


CharlesKrop:
As the title suggests, defining a DataDimension size within a QuantityFactory using an NDSL type causes issues. 

Currently, for things to work properly, data dimensions must be defined using non-NDSL (non-numpy) types. At the same time, however, externals must be defined using NDSL types to ensure that proper precision is retained in calculations.

This causes an issue when you have a parameter which is used as both a dimension size and an external. Right now, using an NDSL type for a QuantityFactory data dimension size causes a cascading issue later on, when that dimension is needed in a stencil. This results in a StencilFactory explosion, and a very unintuitive series of errors which struggle to point to the real - QuantityFactory related - error.

Workarounds are easy enough, but it still should probably be looked at and a proper solution implemented since it is difficult to track the problem back to the original source.

See attached code for repro and example workaround.

[repro_add_dimension_bug.py](https://github.com/user-attachments/files/29815350/repro_add_dimension_bug.py)

romanc:
PR #501 is a stop-gap measure to enforce data dimension sizes of type `int` and issue a clear `TypeError` if not. This - at least - gives NDSL users a clean error message, allowing them to use workarounds like `int()` casting.

In theory, we should be able to allow data dimension sizes of any [`numbers.Integral`](https://docs.python.org/3/library/numbers.html#numbers.Integral) type. Right now, this validates some assumptions in both GT4Py (IRs require data dim sizes to be of type `int`) and DaCe (the `Data` class requires `shape` to be of type `int`). Once those (and potentially others) are updated (might take some time and convincing), we can lift this current limitation in NDSL.

romanc:
In DaCe, according to Tal

> I think that's one of those cases that the stree-based Python frontend fixes (I think) but it's very heavily WIP

Just putting this here as reference.

romanc:
Support added with https://github.com/spcl/dace/issues/2431, https://github.com/GridTools/gt4py/issues/2705, and https://github.com/NOAA-GFDL/NDSL/pull/531. Closing

PR #501, opened 2026-07-10
feat: limit data dimension sizes to python native `int`

romanc:
# Description

Data dimension fields allow users to store additional dimension of a given size. Due to limitations / assumptions in both GT4Py and DaCe, these data dimension have to be of  python native type `int`.

This PR hardens the NDSL side, providing NDSL users with a clear error message as a stop gap measure. Once known, the issue can be worked around easily by an explicit cast when registering the data dimension.

We _should_ be able to allow any `numbers.Integral` type as size for data dimensions. This will require work in GT4Py (IRs will check that `data_dims` is a list/tuple of `int`) and DaCe (the `Data` class currently requires `shape` to be of type `int` or symbolic). Once these (and potentially others) are changed, we can lift this hard limitation in NDSL.

In addition, this PR unearthed two issues when declaring data dimension fields:

1. Declaration would fail when fields were declared with an indent.
2. Declaration would fail when fields were declared over multiple lines (e.g. as part of `declare_and_register()`)

Both have to do with how we automagically find the type name. Case 1 can be solved easily by just stripping any white-space before parsing. This will now just work.

Case 2 is more involved. We could "look ahead" more lines of code if necessary. While the parsing `code_context` is only the current line, we could leverage tools like [linecache](https://docs.python.org/3/library/linecache.html) to "read ahead" additional lines in case of syntax errors (i.e. on a heuristic when the `SyntaxError`'s message contains something like "was never closed"). For now, we just drop a more clear error message.

/cc @FlorianDeconinck FYI

Related issue: #500

## How has this been tested?

Covered by two new test cases.

## Checklist

- [ ] My changes generate no new warnings: N/A


twicki (approved):
Looks good, thanks for improving the user experience

PR #505, opened 2026-07-14
feature: GPU hardware detection for orchestrated backend

romanc:
# Description

This PR pulls out GPU hardware detection of the orchestrated dace backends from PR https://github.com/NOAA-GFDL/NDSL/pull/465.

## How has this been tested?

Mainly tested by running simulations as part of the June push (as part of PR #465). Added very simplistic test cases to cover the basics. It's kinda hard to check for hw detection (because you can't really know the hw you are running on).

## Checklist

### DIFF
--- a/docs/docstrings/dsl/dace/hardware_config.md
+++ b/docs/docstrings/dsl/dace/hardware_config.md
@@ -1,0 +1,12 @@
+# hardware_config
+
+::: dsl.dace.hardware_config
+
+<style>
+/* re-enable the left side navigation bar for this page */
+@media screen and (min-width: 76.1875em) {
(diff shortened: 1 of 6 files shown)


twicki (approved):
With this we currently still have no way of manually overwriting this, correct?

The more we do, the more we need a config of our own. But having this as the default is already better than what we had before

romanc:
> With this we currently still have no way of manually overwriting this, correct?

Correct - there's no way (without code changes) to change these defaults.

> The more we do, the more we need a config of our own. But having this as the default is already better than what we had before

Agreed, the need for an NDSL-level config file gets more and more apparent.This is another good example.

Issue #339, opened 2025-12-09
[Stencils opt] Overuse of `inspect` for the guardrails of Local


FlorianDeconinck:
In order to enforce the `Local`/`LocalState` restriction of usage we are relying heavily on `inspect` which is a very slow call. This call is done _once_ in orchestration and can be ignore. But under stencil operations, it will be done _every single call_ which will lead to performance degradation.

Potential solution:
- cache access after the first one
- unhook guardrails after the first check
- unhook guardrails in "performance" mode

romanc:
Some thought after working on PR #509 (and - arguably - making it worse).

Say we use caching. Caching would look something like "usage of locals is fine _in this context_". However, _in this context_ would still rely on `inspect` to figure out where the local is called from. We'd thus need to store the context as part of the key. And having the context being part of the key means that we'll need to evaluate the context for every lookup. Unhooking the guardrails after the first check would suffer from the same issue.

Given the above, my vote here is for usage of a "performance mode" (ideally configured in an [NDSL-level config file](https://github.com/NOAA-GFDL/NDSL/issues/510) - otherwise another environment variable will do).

PR #509, opened 2026-07-21
feature: guard against usage of locals outside orchestrated code

romanc:
# Description

Locals can only safely be used in orchestrated code paths. This is because in orchestration, we'll put DaCe in charge of memory management of Locals on the cpp/cuda side. There's no back-link to the python memory, so any usage of Locals outside orchestrated code paths will result in backend-depending execution as already anticipated as issue in the original Locals PR #266.

Within `NDSLRuntime`, we already check that usage of Locals is restricted to the class they are defined in. In this PR, we refine this check to forbid access from non-orchestrated methods. To do so, we need to know which methods of `NDSLRuntime` classes are orchestrated. We thus keep taps on those with `_ndsl_orchestrated_methods`, which gets automatically populated as part of `orchestrate()` if the property is detected.

Note: This will only worsen the potential performance problem in stencil mode outlined in https://github.com/NOAA-GFDL/NDSL/issues/339. Since guardrails are still contained to the same `__getattr__()` function, all possible solutions outlined there are still possible. That said, I have picked my favourite solution as mentioned in issue #339.

Related issue: https://github.com/GridTools/gt4py/issues/2665

## How has this been tested?

New / updated unit tests.

## Checklist

### DIFF
--- a/ndsl/dsl/dace/orchestration.py
+++ b/ndsl/dsl/dace/orchestration.py
@@ -755,6 +755,7 @@ def orchestrate(
 ) -> None:
     """
     Orchestrate a method of an object with DaCe.
+
     The method object is patched in place, replacing the original Callable with
     a wrapper that will trigger orchestration at call time.
     If the model configuration doesn't demand orchestration, this won't do anything.
@@ -766,11 +767,21 @@ def orchestrate(
         dace_compiletime_args: list of names of arguments to be flagged has
                                dace.compiletime for orchestration to behave
     """
-    if not config.is_dace_orchestrated():
-        return
+    if hasattr(obj, "_ndsl_orchestrated_methods"):
+        # Automatically register all orchestrated methods of NDSLRuntime classes
+        # to track where Locals can be used.
+        # See __post_init__() of NDSLRuntime.
+        obj._ndsl_orchestrated_methods.append(method_to_orchestrate)
 
     if config is None:
         raise ValueError("DaCe config cannot be None")
+
+    if not config.is_dace_orchestrated():
+        return
+
+    # We have to un-monkey patch the __call__ (from the debugger)
+    if method_to_orchestrate == "__call__" and hasattr(type(obj), "_original__call__"):
+        type(obj).__call__ = type(obj)._original__call__  # type: ignore[method-assign,attr-defined]
 
     if not hasattr(obj, method_to_orchestrate):
         raise RuntimeError(
--- a/ndsl/dsl/ndsl_runtime.py
+++ b/ndsl/dsl/ndsl_runtime.py
@@ -31,6 +31,9 @@ class NDSLRuntime:
         self._stencil_factory = stencil_factory
         # Use this flag to detect that the init wasn't done properly
         self._base_class_was_properly_super_init = True
+        # Used to track where usage of Locals is safe.
+        self._ndsl_orchestrated_methods: list[str] = []
+
         self._optimization_config = optimization_config
 
     def __init_subclass__(cls: type[NDSLRuntime], **kwargs: dict[str, Any]) -> None:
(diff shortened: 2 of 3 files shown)


ndsl/dsl/dace/orchestration.py
```diff
@@ -634,6 +634,12 @@ def orchestrate(
         dace_compiletime_args: list of names of arguments to be flagged has
                                dace.compiletime for orchestration to behave
     """
+    if hasattr(obj, "_ndsl_orchestrated_methods"):
+        # Automatically register all orchestrated methods of NDSLRuntime classes
+        # to track where Locals can be used.
+        # See __post_init__() of NDSLRuntime.
+        obj._ndsl_orchestrated_methods.append(method_to_orchestrate)
```
romanc, line 774: This is somewhat fragile. Another approach would be to have an optional callback as argument of `orchestrate()`, which get called if defined, i.e.

```py
def orchestrate(
    *,
    obj: object,
    config: DaceConfig,
    method_to_orchestrate: str = "__call__",
    dace_compiletime_args: Sequence[str] | None = None,
    pre_orchestration: Callable | None = None,
) -> None:
    """
    Orchestrate a method of an object with DaCe.
    The method object is patched in place, replacing the original Callable with
    a wrapper that will trigger orchestration at call time.
    If the model configuration doesn't demand orchestration, this won't do anything.

    Args:
        obj: object which methods is to be orchestrated
        config: DaceConfig carrying model configuration
        method_to_orchestrate: string representing the name of the method
        dace_compiletime_args: list of names of arguments to be flagged has
                               dace.compiletime for orchestration to behave
        pre_orchestration: If a function is given, it will be called (with this function's
            arguments before the function is orchestrated). 
    """
    if pre_orchestration is not None:
        pre_orchestration(obj, config, method_to_orchestrate, dace_compiletime_args)

    if not config.is_dace_orchestrated():
        pass

    ...
```

That would move the registration to `NDSLRuntime`, which has pros and cons:

- on the plus side, we would have `NDSLRuntime` being in charge of registering orchestrated methods
- on the flip side, users will have to remember to pass this additional argument when the call `orchestrate()` to orchestrate any other method than the call function. We could - of course - remedy this with syntactic sugar in classes deriving from `NDSLRuntime`.

ndsl/dsl/ndsl_runtime.py
```diff
@@ -121,10 +124,7 @@ def check_for_quantity(object_: object) -> None:
             check_for_quantity(self)
 
         # Orchestrate __call__ by default
-        if self._stencil_factory.backend.is_orchestrated() and callable(self):
-            # Do we have to un-monkey patch the __call__
-            if hasattr(type(self), "_original__call__"):
-                type(self).__call__ = type(self)._original__call__  # type: ignore[method-assign,attr-defined]
```
romanc, line 127: Two things here

1. Allow any backend to "orchestrate" such that we can restrict local to non-orchestrated code paths in all backends. I'm happy to debate this; I could also see a version where only orchestrated backends can check usage of locals. I've done it this way to keep the amount of backend-specific behavior to a minimum.
2. Moved the undo of monkey-patching from the debugger into the `orchestrate()`  function as a result of point 1. Since I put it after the early return for non-orchestrated backends, all the conditions of the guarding `if`  statements stay the same.

romanc:
@FlorianDeconinck As outlined above, I'm happy to discuss usage of `_ndsl_orchestrated_methods`. Also, interaction with the new automatic debugger wasn't straight forward. Alt tests seem to pass, but please have a look with the debugger in mind.

FlorianDeconinck (approved):
This works, thank you.

It does highlights some weakness in how we treat runtime restrictions between stencil and orchestrated backends. Debugger is barely a thing in orchestration (only work for the top level call) but plagues us all the way down. On the other hand, `Locals` are spicier in `orchestration`.

We might need to break away from the orchestration idea in implementation and put in a "runtime" concept that defines the critical path of the code and bolt-on the needed changes to make debugger, locals and everything else work.

Baseline: the monkey-patching might be stretch as far as we should.

romanc:
I agree we are starting to reach limits and a ground-up rethink/plan might be appropriate at some point. I'd wait until we have decent performance results and some science code running on NDSL before we go there.

PR #524, opened 2026-08-05
[Feat | Opt II] Simplifying offgrid conditionals pre-merging

FlorianDeconinck:
# Description

Offgrid conditionals drive the flow of code (flags, optional fields) but they block the merging of the axis. In this PR we improve the behavior of inlining conditionals to allow for more merging.

- [Refactor] Pipeline can now call other pipelines
- `CartesianMerger` is now a pipeline
- New `SimplifyConditionals` normalize all constructed in a series of `IF`
- Restrict to `offgrid` conditionals to make sure to not drag cartesian-offset issues
- More topology checks
- Small QOLs to fix debugging in cold cache cases

ToDo:
- [x] Restore ELSE IF and ELSE post merge

Fixes # (issue)

## How has this been tested?

New unit tests

## Checklist

- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (e.g. add new modules to docs/docstrings/)
- [ ] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] New check tests, if applicable, are included

PR #529, opened 2026-08-07
[Fix | Opt II] Exclude top level Labeler from the LocalOptimization scheme

FlorianDeconinck:
# Description

The local optimizer first version code triggers on top level code, which means that the next piece of the optimizer is running on _already_ global optimized code.

We introduce a quick fix to exclude the top level `Labeler` from the `LabelSection` transform which restore original intent.

A larger fix, as written in the code, is to merge the `LocalOptimization` in the larger "pipeline" concept, since this is in essence a localized pipeline.

Note: I introduce another topology function, without a test. I will write all tests in #524 as well as refactoring their API.

## How has this been tested?

Introduced a test for local optimization which test the local optimization is done properly. It doesn't _actually_ reflect the bug, but it checks that local optimization are indeed applied.

## Checklist

### DIFF
--- a/ndsl/dsl/dace/stree/optimizations/common/__init__.py
+++ b/ndsl/dsl/dace/stree/optimizations/common/__init__.py
@@ -3,6 +3,7 @@ from .loops import is_axis_for, is_axis_map, is_cartesian_axis
 from .topology import (
     detect_cycle,
     get_next_node,
+    is_first_node,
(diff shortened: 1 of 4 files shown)


romanc (approved):
Oh wow, good find.

I agree, we can do it like this for now and clean up later as proposed.

PR #531, opened 2026-08-10
feat: Allow data dimension sizes with NDSL integer type

romanc:
# Description

NDSL has an `Int` type that resolves to either a 32- or 64-bit numpy integer type depending on the configured literal precision. Users would like to use this type in the size definition of data dimensions (see related issue).

With DaCe PR https://github.com/spcl/dace/issues/2431 and GT4Py PR https://github.com/GridTools/gt4py/pull/2762 merged, we have the necessary conditions to make this work at the NDSL-level.

I took the liberty to make a `TypeAlias` for `DataDimensions` since it's used in a bunch of places. The newly added `IntegralNumber` (in `ndls.types` is due to the fact that `mypy` doesn't have support for the type hierarchy defined in `numbers` (there are multiple closed `mypy` issues that all basically say "this is hard and won't give much; this is python - don't expect to have super-duper-exact types". So, while I can (and I think we should) use `numbers.Integral` for checking at runtime, using it in type hints isn't really an option because `int` wouldn't be seen as a subtype of `numbers.Integral`. And even if, because `dict` is not covariant, it wouldn't work anyway ... So - long story short - I opted for a type union, which just works as expected.

Related issue: #500

## How has this been tested?

Updated unit test and locally with the repro in the issue.

## Checklist


FlorianDeconinck (approved):
LGTM!

PR #536, opened 2026-08-12
[Fix | Opt Cycle II] Transient strides and lifetime

FlorianDeconinck:
# Description

`RefineTransient` recomputes strides for refined data.
We also force transient to be Persistent on CPU since we have no memory pool. When refining, the transient can be pushed _within_ a loop, and if the lifetime is kept at the default `Scope` it will allocate every time. GPU has a memory pool to alleviate this, CPU doesn't.

Also:
- `Backend.as_layout_map` now knows how to give a map taking a data dimensions into account.

## How has this been tested?

New and adjusted utests

## Checklist

### DIFF
--- a/ndsl/config/backend.py
+++ b/ndsl/config/backend.py
@@ -174,12 +174,25 @@ class Backend:
     def as_safe_for_path(self) -> str:
         return self._humanly_readable.replace(":", "_")
 
-    def as_layout_map(self) -> tuple[int, ...]:
+    def as_layout_map(self, data_dimensions_size: int = 0) -> tuple[int, ...]:
+        """Give backe the layout as a tuple of integer signifying the order of
+        the axis in the strides.
+
+        Args:
+            data_dimensions_size: when > 0 extends the map to include the data_dimensions
+            expected layout
+        """
         loop_order_as_string = self._loop_order.value
-        return tuple(
+        cartesian = [
             len(loop_order_as_string) - 1 - loop_order_as_string.index(axis)
             for axis in "IJK"
+        ]
+        data_dimensions = (
+            [ddim + len(cartesian) for ddim in range(data_dimensions_size)]
+            if data_dimensions_size > 0
+            else []
         )
+        return tuple(cartesian + data_dimensions)
 
     def is_orchestrated(self) -> bool:
         return self._strategy == BackendStrategy.ORCHESTRATION
--- a/ndsl/config/backend.py
+++ b/ndsl/config/backend.py
@@ -174,8 +174,8 @@ class Backend:
     def as_safe_for_path(self) -> str:
         return self._humanly_readable.replace(":", "_")
 
-    def as_layout_map(self, data_dimensions_size: int = 0) -> tuple[int, ...]:
-        """Give backe the layout as a tuple of integer signifying the order of
+    def as_layout_map(self, *, data_dimensions_size: int = 0) -> tuple[int, ...]:
+        """Give back the layout as a tuple of integer signifying the order of
         the axis in the strides.
 
         Args:
@@ -187,11 +187,7 @@ class Backend:
             len(loop_order_as_string) - 1 - loop_order_as_string.index(axis)
             for axis in "IJK"
         ]
-        data_dimensions = (
-            [ddim + len(cartesian) for ddim in range(data_dimensions_size)]
-            if data_dimensions_size > 0
-            else []
-        )
+        data_dimensions = [ddim + len(cartesian) for ddim in range(data_dimensions_size)]
         return tuple(cartesian + data_dimensions)
 
     def is_orchestrated(self) -> bool:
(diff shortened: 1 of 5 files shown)


ndsl/config/backend.py
```diff
@@ -174,12 +174,25 @@ def as_humanly_readable(self) -> str:
...
         loop_order_as_string = self._loop_order.value
-        return tuple(
+        cartesian = [
             len(loop_order_as_string) - 1 - loop_order_as_string.index(axis)
             for axis in "IJK"
+        ]
+        data_dimensions = (
+            [ddim + len(cartesian) for ddim in range(data_dimensions_size)]
+            if data_dimensions_size > 0
+            else []
```
romanc: I think the `if` / `else`  is not necessary. According to a python repl, the list creation will behave the same without the `if/else`:

````none
>>> size = 0
>>> [ddim for ddim in range(size)]
[]
>>> [ddim for ddim in range(-1)]
[]
```

ndsl/config/backend.py
```diff
@@ -174,12 +174,25 @@ def as_humanly_readable(self) -> str:
     def as_safe_for_path(self) -> str:
         return self._humanly_readable.replace(":", "_")
 
-    def as_layout_map(self) -> tuple[int, ...]:
+    def as_layout_map(self, data_dimensions_size: int = 0) -> tuple[int, ...]:
+        """Give backe the layout as a tuple of integer signifying the order of
```
romanc:
```suggestion
        """Give back the layout as a tuple of integers signifying the order of
```

ndsl/dsl/dace/stree/optimizations/refine_transients.py
```diff
@@ -59,7 +60,18 @@ def _reduce_cartesian_axis_size_to_1(
...
+    # ⚠️ The below ddim calculation only works because we _know_ it's a 3D field
+    #    We need to somehow carry the information of the cartesian access here to go to 3D
+    transient_data.set_strides_from_layout(
+        *backend.as_layout_map(data_dimensions_size=len(new_shape) - 3)
+    )
+
+    # CPU doesn't carry a memory pool - therefore any allocation will end
+    # up on the scope it is needed. Post refine, this can mean allocation every
+    # loop iteration. We push the transient on Persistent to make sure this does not happen.
+    if not backend.is_gpu_backend():
```
romanc: I'd prefer if we added `Backend.is_cpu_backend()`  just to be clear that we are expecting a CPU backend here.

ndsl/config/backend.py
```diff
@@ -174,12 +174,25 @@ def as_humanly_readable(self) -> str:
     def as_safe_for_path(self) -> str:
         return self._humanly_readable.replace(":", "_")
 
-    def as_layout_map(self) -> tuple[int, ...]:
+    def as_layout_map(self, data_dimensions_size: int = 0) -> tuple[int, ...]:
```
romanc: how about 

```suggestion
    def as_layout_map(self, *, data_dimensions_size: int = 0) -> tuple[int, ...]:
```

such that we forbid usage of `backend.as_layout_map(2)`  where just `2` reads very ambiguous to me.

romanc (approved):
Looks good in general. Some nitpicks/suggestions inline. Take what you think makes sense.

I'd propose to add the following test case to `tests/config/test_backend.py`

```py
def test_backup_as_layout_map() -> None:
    ijk_backend = Backend("st:python:cpu:IJK")
    assert ijk_backend.as_layout_map() == (2,1,0)
    assert ijk_backend.as_layout_map(data_dimensions_size=3) == (2,1,0,3,4,5)
    assert ijk_backend.as_layout_map(data_dimensions_size=0) == (2,1,0)
    assert ijk_backend.as_layout_map(data_dimensions_size=-1) == (2,1,0)
```

that should cover typical cases and some unexpected/edge cases

ndsl/config/backend.py
```diff
@@ -174,12 +174,25 @@ def as_humanly_readable(self) -> str:
...
+            data_dimensions_size: when > 0 extends the map to include the data_dimensions
+            expected layout
+        """
         loop_order_as_string = self._loop_order.value
-        return tuple(
+        cartesian = [
             len(loop_order_as_string) - 1 - loop_order_as_string.index(axis)
             for axis in "IJK"
+        ]
+        data_dimensions = (
```
twicki: Thanks for adding this!

twicki (approved):

FlorianDeconinck:
> Looks good in general. Some nitpicks/suggestions inline. Take what you think makes sense.
> 
> I'd propose to add the following test case to `tests/config/test_backend.py`
> 
> ```python
> def test_backup_as_layout_map() -> None:
>     ijk_backend = Backend("st:python:cpu:IJK")
>     assert ijk_backend.as_layout_map() == (2,1,0)
>     assert ijk_backend.as_layout_map(data_dimensions_size=3) == (2,1,0,3,4,5)
>     assert ijk_backend.as_layout_map(data_dimensions_size=0) == (2,1,0)
>     assert ijk_backend.as_layout_map(data_dimensions_size=-1) == (2,1,0)
> ```
> 
> that should cover typical cases and some unexpected/edge cases

Added that and extended it to also cover another layout backend to make sure we don't regress into a space where every layout answer the same

PR #537, opened 2026-08-13
[Feat | Lang] Introducing `ConstantQuantity`

FlorianDeconinck:
# Description

With `Local` being a new symbolic way to flag for "temporary" memory to be optimized later. We also have implemented a warning showing any memory living on a `NDSLRuntime` and that is _not_ flagged as a `Local`.

We are still seeing the warnings pop in and some are pointing to a missing use case: constants buffer. E.g. the following damping coefficient in the column calculated during `__init__` and used afterwards:

```python
class DelnFlux(NDSLRuntime):
    def __init__(...):
        [...]
        self._damp = calc_damp(
            damp_c=damp_c, da_min=damping_coefficients.da_min, nord=nord_col
        )
```

We propose a `ConstantQuantity` that forbids writing into it past it's initialization and flags the usage so that following optimization can be done (not part of this PR).

## How has this been tested?

New utest.

## Checklist

### DIFF
--- a/ndsl/dsl/ndsl_runtime.py
+++ b/ndsl/dsl/ndsl_runtime.py
@@ -2,8 +2,10 @@ from __future__ import annotations
 
 import inspect
 import warnings
-from collections.abc import Callable
-from typing import Any, Sequence
+from collections.abc import Callable, Sequence
+from typing import Any
+
+import numpy as np
 
 from ndsl import OptimizationConfig
 from ndsl.debug import get_debugger
@@ -11,7 +13,12 @@ from ndsl.dsl.dace.orchestration import orchestrate
 from ndsl.dsl.stencil import StencilFactory
 from ndsl.dsl.typing import Float
 from ndsl.initialization.allocator import QuantityFactory
+from ndsl.optional_imports import cupy
 from ndsl.quantity import Local, Quantity
+from ndsl.quantity.local import ConstantQuantity
+
+if cupy is None:
+    import numpy as cupy
 
 _TOP_LEVEL: object | None = None
 
@@ -192,5 +199,32 @@ class NDSLRuntime:
             origin=quantity.origin,
             extent=quantity.extent,
             backend=quantity.backend,
+            allow_mismatch_float_precision=allow_mismatch_float_precision,
+        )
+
+    def make_constant_quantity(
+        self,
+        data: np.ndarray | cupy.ndarray | Quantity,
+        quantity_factory: QuantityFactory,
+        dims: Sequence[str],
+        dtype: type = Float,
+        units: str = "unspecified",
+        *,
+        allow_mismatch_float_precision: bool = False,
+    ) -> ConstantQuantity:
+        constant = quantity_factory.zeros(
+            dims,
+            units,
+            dtype,
+            allow_mismatch_float_precision=allow_mismatch_float_precision,
+        )
+        constant.field[:] = data[:]
+        return ConstantQuantity(
+            data=constant._data,
+            dims=constant.dims,
+            units=constant.units,
+            origin=constant.origin,
+            extent=constant.extent,
+            backend=constant.backend,
             allow_mismatch_float_precision=allow_mismatch_float_precision,
         )
--- a/ndsl/quantity/local.py
+++ b/ndsl/quantity/local.py
@@ -54,3 +54,34 @@ class Local(Quantity):
             f"    units={self.units},\n    origin={self.origin},\n"
             f"    extent={self.extent}\n)"
         )
+
+
+class ConstantQuantity(Quantity):
+    """ConstantQuantity is a Quantity that cannot be written to outside of it's allocation
+    (notably with the `NDSLRuntime.make_constant_quantity` method)."""
+
+    def __init__(
+        self,
+        data: np.ndarray | cupy.ndarray | Quantity,
+        dims: Sequence[str],
+        units: str,
+        *,
+        backend: Backend,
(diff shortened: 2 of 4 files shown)


ndsl/dsl/ndsl_runtime.py
```diff
@@ -194,3 +201,30 @@ def make_local(
...
+        *,
+        allow_mismatch_float_precision: bool = False,
+    ) -> ConstantQuantity:
+        constant = quantity_factory.zeros(
+            dims,
+            units,
+            dtype,
+            allow_mismatch_float_precision=allow_mismatch_float_precision,
+        )
+        constant.field[:] = data[:]
```
romanc, line 221: this reads weird ... why restrict to `.field`?

```suggestion
        constant[:] = data[:]
```

ndsl/quantity/local.py
```diff
@@ -54,3 +54,34 @@ def __repr__(self) -> str:
             f"    units={self.units},\n    origin={self.origin},\n"
             f"    extent={self.extent}\n)"
         )
+
+
+class ConstantQuantity(Quantity):
```
romanc, line 59: I think this warrants a separate file

ndsl/quantity/local.py
```diff
@@ -54,3 +54,34 @@ def __repr__(self) -> str:
...
+            origin=origin,
+            extent=extent,
+            allow_mismatch_float_precision=allow_mismatch_float_precision,
+            backend=backend,
+        )
+
+    def __setitem__(self, subscript: Any, value: Any) -> None:
+        raise RuntimeError(
+            "ConstantQuantity cannot be written too after initialization"
+        )
```
romanc, line 87: we might need more guardrails here. Looking at `Quantity`, I think we'll also need to access to `.field`,  and `swap_buffer`. Not sure about `.view`  and `.sel(...)`

romanc (commented):
My big question here is: what is the optimization (potential) that you see given that we had `ConstantQuantities`?

ndsl/dsl/ndsl_runtime.py
FlorianDeconinck: Because 90% of the time people have data for the domain of their field, not the halos. If I am building, like in the example, a set of coefficient, I am thinking of the domain.

FlorianDeconinck:
> My big question here is: what is the optimization (potential) that you see given that we had `ConstantQuantities`?

This serves two purposes:
- optimization: mostly inlining of values, also in limited area a better selection of the memory residency. This would apply even more for large tables
- semantic check: we can keep our frontend checks for quantities that are on a `NDSLRuntime` with proper semantic tagging (e.g. I won't write into this but it's not a Local per se)

My main problem with this is that it's too adjacent to the concept of constant and global table.

ndsl/dsl/ndsl_runtime.py
romanc: I agree and I think this should be reflected in the Quantity (constant or not); i.e. in your example, imo the Quantity should have no halos and then no restriction to `.field` is necessary.

FlorianDeconinck: But `halos` is (supposedly) a transparent concept for users. The vast majority of modelers that are not working on dynamics don't know what a halo is. So you see it as a restriction because you are aware of the memory, most users would calculate on a C24 a 24x24 field, then this would fail because the Quantity is actually 30x30 or even 31x31 due to our weird extra point.

FlorianDeconinck: Ha I get your point, the `ConstantQuantity` should have no halo by default

romanc: Exactly, since everything is constant, there's no need for a halo because the values never change. With this new class, we can also force not to have the weird extra point from the get go. This might (or not) get awkward with the `QuantityFactory`, but since we're only allocating (the quantity) from `NDSLRuntime`, we should be able to have some sort of specialized `QuantityFactory` there.

romanc:
> This serves two purposes:
> 
>     * optimization: mostly inlining of values, also in limited area a better selection of the memory residency. This would apply even more for large tables
> 
>     * semantic check: we can keep our frontend checks for quantities that are on a `NDSLRuntime` with proper semantic tagging (e.g. I won't write into this but it's not a Local per se)

I think both points make sense. For a moment, I was afraid this is all just to suppress the warnings you talk about in the description.

> My main problem with this is that it's too adjacent to the concept of constant and global table.

You mean in terms of wording? Maybe it helps to have a custom example, i.e. how would the damping coefficients in `DelnFlux` change if we add `ConstantQuantities`? What are the differences/similarities between `ConstantQuantities` and global tables?

FlorianDeconinck:
> What are the differences/similarities between ConstantQuantities and global tables?

One is a NDSL system, one is a GT4Py system - they are largely similar. `GlobalTables` doesn't enforce the read-only, but from a user point of view it feels weird to make the difference

romanc:
> > What are the differences/similarities between ConstantQuantities and global tables?
> 
> One is a NDSL system, one is a GT4Py system - they are largely similar. `GlobalTables` doesn't enforce the read-only, but from a user point of view it feels weird to make the difference

So both (global tables and constant quantities) are constant, off-grid data? Or do we also foresee usage of grid-based constant quantities?

Is the "table" terminology something we like in NDSL?

- if yes, we could make it a first-class citizen with the meaning of "constant, off-grid data"
- if not, we could "hide it" and just offer `Quantity` / `ConstantQuantity` at the NDSL-level

PR #538, opened 2026-08-14
[Devops] Move NDSL to `uv` 

FlorianDeconinck:
# Description

This PR moves NDSL to use `uv` as a package manager. CI, README and [dev setup guide](https://github.com/GEOS-ESM/SMT-Nebulae#dev-setup-guide) have been updated to reflect the changes.

In this PR we remove the submodules in favor of pinned git-dependencies to `gt4py` and `dace`. This simplifies user installation considerably (no need to know/learn submodules). In the back, we keep all the flexibility of pointing to (almost) arbitrary versions of `gt4py`  and `dace`.

We have future PRs in mind to move the linting to `ruff`. That's outside the scope of this PR.

## How has this been tested?

CI shall bless the new installs.

## Checklist


romanc:
> * adding `ruff.toml` to skip suggestions that conflict with other tools (isort) or are deliberate decision by the team

if we are going to `ruff` (I think it's a good idea), I would just use `ruff`  and replace `isort`, `black` and `flake8`.

FlorianDeconinck:
> > * adding `ruff.toml` to skip suggestions that conflict with other tools (isort) or are deliberate decision by the team
> 
> if we are going to `ruff` (I think it's a good idea), I would just use `ruff` and replace `isort`, `black` and `flake8`.

I like that, but I haven't really looked if it's going to re-lint the code base too much (e.g. we need to do it with some care so we skip some merging pain)

romanc:
> > > * adding `ruff.toml` to skip suggestions that conflict with other tools (isort) or are deliberate decision by the team
> > 
> > 
> > if we are going to `ruff` (I think it's a good idea), I would just use `ruff` and replace `isort`, `black` and `flake8`.
> 
> I like that, but I haven't really looked if it's going to re-lint the code base too much (e.g. we need to do it with some care so we skip some merging pain)

I see - makes sense. Let's do one at a time then.

romanc:
okay, basic version is working for just inside NDSL when you have the whole stack on `uv`.

the complicated thing here is that `uv.sources`  is only visible to `uv`, so if someone (e.g. pyFV3) is using something else than `uv` (e.g. plain `pip`), then they don't see the version in `uv.sources`, i.e. `gt4py`  and `dace`  are back to the last released version on pip. That's why tests are failing on `gtscript.enum` because that's not (yet) in gt4py  version 1.2.0.

romanc:
optimized "dev dependencies" such that for linting only the linting tools get installed and for docs build only the docs get installed. This part works nice. To avoid the repetitive arguments to `uv run` (e.g. `--frozen`), we should set the following env variables (workflow-wide if possible)

- `UV_FROZEN`: Sync without updating the `uv.lock` file (we check that `uv.lock` is up-to-date in the linting workflow. all other workflows should really just install versions specified in the lock file to ensure repeatable builds)
- `UV_NO_DEV`: Disable the development dependency group (we enable specific groups when we need them)
- `UV_NO_EDITABLE`: Install the project as non-editable (there's really no need to in CI workflows)

romanc:
Alright - this is just missing a blurb in SMT on how to setup "the complicated" dev setup.

romanc (approved):
I am happy with this. It's up for show/review now. Please ask any questions you might have.

FlorianDeconinck (commented):
LGTM - let's have this discussed and reviewed by @oelbert / @fmalatino 

The main change here is:
- users needs to go to the `uv` toolchain
- `gt4py`/`dace` are no longer `submodules` and developers will have to reset their working stack as per docs

fmalatino (approved):
Really glad to see these changes. Has a test build with pyFMS using the new build backend?

romanc:
> The main change here is:
> 
>  * users needs to go to the `uv` toolchain

Just to clarify this point: There's currently nothing `uv`-specific in the configuration. If users wish, they can keep installing with plain `pip`. For example, the translate tests of `pyFV3`, `pySHiELD` and `pace`  currently do so and they work as before.

We are nudging users (and developers) towards `uv` because we (at least I) think it's nice to work with, but - as said - all configuration in `pypropject.toml`  remains spec compliant and thus any dependency manger (`uv`, `pip`, `poetry`, ...) will work just fine.

romanc:
> Has a test build with pyFMS using the new build backend?

Yes. We run pyFMS-tests as part of [.github/workflows/unit_tests.yml](https://github.com/FlorianDeconinck/NDSL/blob/devops/uv_pyproject/.github/workflows/unit_tests.yaml) in a check called `NDSL unit tests / Python 3.13 (pyFMS)`.

image

The log of the latest run of this workflow can be seen [here](https://github.com/NOAA-GFDL/NDSL/actions/runs/32712167695/job/97385716967?pr=538).

fmalatino:
> > Has a test build with pyFMS using the new build backend?
> 
> 
> 
> Yes. We run pyFMS-tests as part of [.github/workflows/unit_tests.yml](https://github.com/FlorianDeconinck/NDSL/blob/devops/uv_pyproject/.github/workflows/unit_tests.yaml) in a check called `NDSL unit tests / Python 3.13 (pyFMS)`.
> 
> 
> 
> image
> 
> 
> 
> The log of the latest run of this workflow can be seen [here](https://github.com/NOAA-GFDL/NDSL/actions/runs/32712167695/job/97385716967?pr=538). 
> 
> 

Thanks! I forgot about this

twicki (approved):
This is a great step!

romanc:
@oelbert what do you think?

PR #543, opened 2026-08-18
[Refactor | Feature] Introducing `DaceExecutable` blind record and replay + refactor the orchestration pipeline

FlorianDeconinck:
# Description

We propose here a wide refactor to bring `DaceExecutable` at the center of the runtime strategy of orchestration. This leads to:
- the `orchestrate` file is now responsible only from the early orchestration steps, moving the full program optimizer across the code
- the new `builder` directory gather all steps to generate a `DaceExecutable` from code

This also brings a `serialize` and `replay` function to `DaceExecutable` which will allow to save the code + data at runtime and replay it blindly without having the original code.

Also:
- `ruff` helped me modernize bits of code

## How has this been tested?

Unit tests + new test for replay

## Checklist


romanc (commented):
From what you explained earlier, this is starting to look nice. I wonder wonder two things (this is getting bigger and bigger):

For one, should we move all of `dsl/dace/stree` under `builder`, i.e.

```none
dsl/
  caches/
  dace/
    builder/
      sdfg/
      stree/      <- this guy is new
      builder.py
      optimize.py
      parse.py
...
```

such that `optimize.py` can get the pipeline and stree optimizations from a "lower" (in the tree) directory rather than a sibling? Are we using the `stree/` directory in some other place?

And the second one; what to do with the file `ndsl/dsl/dace/build.py`. That looks a bit strange (in terms of naming) in my eyes next to a folder that's called `builder/`. The file contains

- `unblock_waiting_tiles()`  unused within NDSL
- `build_info_filepath()` used in the same file inside `write_build_info()`  and `get_sdfg_path()`
- `write_build_info()`: now used from "lower" `dsl/dace/builder/optimize.py`
- `get_sdfg_path()`: now used from "lower" `dsl/dace/builder/{optimize,parse}.py`
- `set_distributed_caches()`  which is used in the `DaceConfig`

To me, it feels like that file should also move into `dsl/dace/builder/` (except for `set_distributed_caches()`, which can probably just be a private method of the `DaceConfig`. That said, I've no clue how much any of those functions are used outside NDSL in places where we import what we need because we know it's there.

FlorianDeconinck:
> From what you explained earlier, this is starting to look nice. I wonder wonder two things (this is getting bigger and bigger):
> 
> For one, should we move all of `dsl/dace/stree` under `builder`, i.e.
> 
> ```
> dsl/
>   caches/
>   dace/
>     builder/
>       sdfg/
>       stree/      <- this guy is new
>       builder.py
>       optimize.py
>       parse.py
> ...
> ```
> 
> such that `optimize.py` can get the pipeline and stree optimizations from a "lower" (in the tree) directory rather than a sibling? Are we using the `stree/` directory in some other place?
> 
> And the second one; what to do with the file `ndsl/dsl/dace/build.py`. That looks a bit strange (in terms of naming) in my eyes next to a folder that's called `builder/`. The file contains
> 
> * `unblock_waiting_tiles()`  unused within NDSL
> * `build_info_filepath()` used in the same file inside `write_build_info()`  and `get_sdfg_path()`
> * `write_build_info()`: now used from "lower" `dsl/dace/builder/optimize.py`
> * `get_sdfg_path()`: now used from "lower" `dsl/dace/builder/{optimize,parse}.py`
> * `set_distributed_caches()`  which is used in the `DaceConfig`
> 
> To me, it feels like that file should also move into `dsl/dace/builder/` (except for `set_distributed_caches()`, which can probably just be a private method of the `DaceConfig`. That said, I've no clue how much any of those functions are used outside NDSL in places where we import what we need because we know it's there.

Agreed.

`stree` went down to `builder`.
The `build.py` was dismantled with code either going closer to the source code for now (this part needs refactor) or going to `dsl.caches` with other caching utilities.

More can be done around the caching and it's utilities to make or retrieve SDFG and cache directories, but at least it cleans it a bit.

FlorianDeconinck:
:bricks: To be merged post 2026.08.00 release :bricks:

ndsl/dsl/caches/__init__.py
```diff
@@ -1,7 +1,9 @@
-from .cache_location import identify_code_path
+from .cache_location import get_cache_directory, get_cache_fullpath, identify_code_path
 from .codepath import FV3CodePath
 
 __all__ = [
-    "identify_code_path",
     "FV3CodePath",
+    "get_cache_directory",
+    "get_cache_fullpath",
+    "identify_code_path",
```
romanc, line 8: should `get_sdfg_path` (from the new `dsl/caches/dace.py`) also be exported here?

ndsl/dsl/caches/dace.py
```diff
@@ -0,0 +1,69 @@
+from ndsl import ndsl_log
+from ndsl.config import Backend
+from ndsl.dsl.caches.cache_location import get_cache_fullpath
+from ndsl.dsl.dace.dace_config import DaceConfig, DaCeOrchestration
+
+DACE_BUILD_INFO_FILENAME = "build_info.txt"
+"""File dropped alongside .dacecache and keeping track of NDSL configuration nformation
+that led to this build."""
```
romanc:
```suggestion
"""File dropped inside the `.dacecache`, which keeps track of NDSL configuration information
that led to this build."""
```

ndsl/dsl/caches/dace.py
```diff
@@ -0,0 +1,69 @@
...
+) -> str | None:
+    """Utility to make an SDFG path from the qualified program name or it's direct path to .sdfg
+
+    Args:
+        daceprog_name: qualified name in the form module_qualname if module is not locals
+        sdfg_file_path: absolute path to a .sdfg file
+    """
+    import os
+
+    # TODO: check DaceConfig for cache.strategy == name
```
romanc: still to do or should this be a follow-up issue?

ndsl/dsl/caches/dace.py
```diff
@@ -0,0 +1,69 @@
...
+    # Guarding against bad usage of this function
+    if not override_run_only and config.get_orchestrate() != DaCeOrchestration.Run:
+        return None
```
romanc: if we get here, can this ever be valid usage of this function? If not, how about we raise an error instead? If we do, I think this function always returns the sdfg path, which means that we can change the return type and remove the `| None` . And that - in turn - would simplify downstream code (e.g. the call inside `get_dace_executable()`).

ndsl/dsl/caches/dace.py
```diff
@@ -0,0 +1,69 @@
...
+    config: DaceConfig,
+    sdfg_file_path: str | None = None,
+    override_run_only: bool = False,
+) -> str | None:
+    """Utility to make an SDFG path from the qualified program name or it's direct path to .sdfg
+
+    Args:
+        daceprog_name: qualified name in the form module_qualname if module is not locals
+        sdfg_file_path: absolute path to a .sdfg file
+    """
```
romanc: Couple things here

1. As far as I can see, `sdfg_file_path` is never used. I propose to remove this argument (and remove the associated code inside the function)
2. I think it would be nice to force `override_run_only`  to be a keyword argument
3. As mentioned inline, I think we should not return `None` and instead raise an error directly here
4. It might help to rely in `pathlib.Path` objects instead of passing around strings as file paths


```suggestion
def get_sdfg_path(
    daceprog_name: str,
    config: DaceConfig,
    *,
    override_run_only: bool = False,
) -> Path:
    """Utility to make an SDFG path from the qualified program name and the DaceConfig.

    Args:
        daceprog_name: qualified name in the form module_qualname if module is not locals
        config: ...
        override_run_only: ...
    """
```

ndsl/dsl/caches/dace.py
```diff
@@ -0,0 +1,69 @@
...
+    # Case of loading a precompiled .so - lookup using GT_CACHE
+    cache_fullpath = get_cache_fullpath(config.code_path)
+    sdfg_dir_path = f"{cache_fullpath}/dacecache/{daceprog_name}"
+    if not os.path.isdir(sdfg_dir_path):
+        raise RuntimeError(f"Precompiled SDFG is missing at {sdfg_dir_path}")
+
+    # Check layout in build time matches layout now
+    import ast
+
+    with open(f"{sdfg_dir_path}/{DACE_BUILD_INFO_FILENAME}") as build_info_file:
```
romanc: For writing `build_info.txt`, there's a function (`_write_build_info()`  in `optimize.py`). I think reading `build_info.txt` also warrants a function. And in a perfect world, these two functions (reading & writing the file) are close to each other. How about we move them to `ndsl/dsl/caches/build_info.py`?

ndsl/dsl/dace/builder/builder.py
```diff
@@ -0,0 +1,84 @@
+from __future__ import annotations
+
+import copy
+import sys
+from typing import Any
+
+from dace import SDFG
+from dace.frontend.python.parser import DaceProgram
+
+import ndsl.dsl.dace.replacements  # noqa # We load in the DaCe replacements
```
romanc: We seem to load the dace replacement repo in multiple places now. Is that by design?

- `orchestration.py`
- `builder.py`
- `optimize.py`
- `parse.py`

all seem to import it now

ndsl/dsl/dace/builder/builder.py
```diff
@@ -0,0 +1,84 @@
+from __future__ import annotations
```
romanc: Is this needed? I thought at some point we moved away from using future annotations

ndsl/dsl/dace/builder/optimize.py
```diff
@@ -0,0 +1,434 @@
...
+    args: Any,
+    kwargs: Any,
+) -> CompiledSDFG | None:
+    """Optimize and compile the SFDG (creating the .daceache and with the source and dynamic library)
+    from the parsed SDFG (e.g. python code + gt4py stencils) when on compiling rank.
+    If called on a non-compiling rank, the functions return None."""
+    is_compiling = DEACTIVATE_DISTRIBUTED_DACE_COMPILE or config.do_compile
+
+    if not is_compiling:
+        return None
```
romanc: just some whitespace changes to make this easier parse-able by human eyes

```suggestion
    """Optimize and compile the SFDG (creating the .daceache and with the source and dynamic library)
    from the parsed SDFG (e.g. python code + gt4py stencils) when on compiling rank.
    
    If called on a non-compiling rank, the functions returns None.
    """
    is_compiling = DEACTIVATE_DISTRIBUTED_DACE_COMPILE or config.do_compile
    if not is_compiling:
        return None
```

ndsl/dsl/dace/builder/builder.py
```diff
@@ -0,0 +1,84 @@
...
+    """Get DaceExecutable by either building or fetching it from cache.
+
+    This function is multi-process safe."""
+
+    mode = config.get_orchestrate()
+    if (
+        mode in [DaCeOrchestration.Build, DaCeOrchestration.BuildAndRun]
+        and dace_program not in DACE_EXECUTABLE_CACHE  # already cached
+    ):
+        parsed_sdfg = copy.copy(sdfg)
```
romanc: maybe add a comment that we save the original sdfg here for later storage in the DaceExecutable ...

or how about we rename the argument `sdfg` -> `parsed_sdfg`  and then this line could read

```suggestion
        original_sdfg = copy.copy(parsed_sdfg)
```

ndsl/dsl/dace/builder/optimize.py
```diff
@@ -0,0 +1,434 @@
...
+
+    This will raise an assertion if cupy is not installed.
+    """
+    assert cp is not None
+    for i, data in enumerate(host_data):
+        if isinstance(data, cp.ndarray):
+            host_data[i] = cp.asarray(data)
+
+
+def _download_results_from_dace(
```
romanc: vscode tells me this function is unused. maybe because it's now in `DaceExecutable`?

ndsl/dsl/dace/builder/optimize.py
```diff
@@ -0,0 +1,434 @@
...
+        return GPUPipeline(
+            config, backend, passes=passes, cache_directory=cache_directory
+        )
+
+    raise ValueError(
+        f"Unknown device type `{device_type}`, expected {DeviceType.CPU} or {DeviceType.GPU}."
+    )
+
+
+def _write_build_info(
```
romanc: How about we move reading / writing `build_info.txt` into a separate module? I think `ndsl/dsl/caches/build_info.py` could be a nice place.

See https://github.com/NOAA-GFDL/NDSL/pull/543/changes#r3819049077.

ndsl/dsl/dace/builder/builder.py
```diff
@@ -0,0 +1,84 @@
...
+        compiled_sdfg = optimize_full_program_sdfg(
+            sdfg, config, optimization_config, args, kwargs
+        )
+
+        MPI.COMM_WORLD.Barrier()
+
+        if not compiled_sdfg:
```
romanc:
```suggestion
        if compiled_sdfg is None:
```

helps the static type checkers to narrow down the type of `compiled_sdfg`

ndsl/dsl/dace/builder/builder.py
```diff
@@ -0,0 +1,84 @@
...
+            ndsl_log.info(
+                f"{DaCeProgress.default_prefix(mode)} Rank is not compiling. "
+                "Waiting for compilation to end on all other ranks..."
```
romanc, line 79: If I understand correctly, this will only print _after_ the compilation ended because of the mpi barrier before. I think if we want to print this before/during compiling, I think the code flow should be something like this

```py
compiled_sdfg = optimize_full_program_sdfg(...)

if compiled_sdfg is None:
    ndsl_log.info("I'm waiting for the compiling rank ...")

MPI.COMM_WORLD.Barrier() # compiling rank is done after that barrier

if compiled_sdfg is None:
    # load sdfg on non-compiling ranks
    ...

MPI.COMM_WORLD.Barrier() # sdfg is loaded on all ranks after this barrier
```

ndsl/dsl/dace/builder/optimize.py
```diff
@@ -0,0 +1,434 @@
+from __future__ import annotations
```
romanc: Is this needed? I thought at some point we moved away from using future annotations

ndsl/dsl/dace/builder/optimize.py
```diff
@@ -0,0 +1,434 @@
...
+from dace.dtypes import ScheduleType
+from dace.dtypes import StorageType as DaceStorageType
+from dace.sdfg.analysis.schedule_tree import treenodes as tn
+from dace.transformation.auto.auto_optimize import make_transients_persistent
+from dace.transformation.dataflow import MapCollapse, MapExpansion
+from dace.transformation.dataflow.add_threadblock_map import AddThreadBlockMap
+from dace.transformation.helpers import get_parent_map
+from gt4py import storage as gt_storage
+
+import ndsl.dsl.dace.replacements  # noqa # We load in the DaCe replacements
```
romanc: We seem to load the dace replacement repo in multiple places now. Is that by design?

- `orchestration.py`
- `builder.py`
- `optimize.py`
- `parse.py`

all seem to import it now

ndsl/dsl/dace/builder/parse.py
```diff
@@ -0,0 +1,91 @@
+from __future__ import annotations
+
+import os
+from typing import Any
+
+from dace import SDFG, CompiledSDFG
+from dace.frontend.python.parser import DaceProgram
+
+import ndsl.dsl.dace.replacements  # noqa # We load in the DaCe replacements
```
romanc: We seem to load the dace replacement repo in multiple places now. Is that by design?

- `orchestration.py`
- `builder.py`
- `optimize.py`
- `parse.py`

all seem to import it now

ndsl/dsl/dace/builder/parse.py
```diff
@@ -0,0 +1,91 @@
+from __future__ import annotations
```
romanc: Is this needed? I thought at some point we moved away from using future annotations

ndsl/dsl/dace/builder/optimize.py
```diff
@@ -0,0 +1,434 @@
...
+    """Optimize and compile the SFDG (creating the .daceache and with the source and dynamic library)
+    from the parsed SDFG (e.g. python code + gt4py stencils) when on compiling rank.
+    If called on a non-compiling rank, the functions return None."""
+    is_compiling = DEACTIVATE_DISTRIBUTED_DACE_COMPILE or config.do_compile
+
+    if not is_compiling:
+        return None
+
+    device_type = DaceDeviceType.GPU if config.is_gpu_backend() else DaceDeviceType.CPU
+    backend_name = config.get_backend()
```
romanc: this is the full backend, not just the name. maybe do a quick rename while we are here.
```suggestion
    backend = config.get_backend()
```

ndsl/dsl/dace/builder/parse.py
```diff
@@ -0,0 +1,91 @@
...
+
+def parse_sdfg(
+    dace_program: DaceProgram,
+    config: DaceConfig,
+    optimization: OptimizationConfig | None,
+    *args: Any,
+    **kwargs: Any,
+) -> SDFG | CompiledSDFG | None:
+    """Return an SDFG depending on cache existence.
+    Either parses, load a .sdfg or load .so (as a compiled sdfg)
```
romanc: There's one thing, I don't quite understand about this function:

If `dace_program`  is in the `DACE_EXECUTABLE_CACHE`, then return it. If we load a pre-compiled SDFG, we put in the cache and return the loaded one. I understand these two branches. The other two branches (`dace_program.to_sdfg()` and `dace_program.load_sdfg()`), we don't populate the cache. Is this by design? Or do we have stale branches here?

ndsl/dsl/dace/builder/optimize.py
```diff
@@ -0,0 +1,434 @@
...
+    DaCeProgress,
+    memory_static_analysis,
+    report_memory_static_analysis,
+)
+from ndsl.optional_imports import cupy as cp
+
+_INTERNAL__SCHEDULE_TREE_OPTIMIZATION_PASSES: list[tn.ScheduleNodeVisitor] | None = None
+
+
+def _upload_to_device(host_data: list) -> None:
```
romanc: Duplicated with `DaceExecutable`

ndsl/dsl/dace/dace_executable.py
```diff
@@ -1,29 +1,208 @@
...
 
+DaceExecutables = dict[DaceProgram, "DaceExecutable"]
+DACE_EXECUTABLE_CACHE: DaceExecutables = {}
+
+_BUNDLE_DIRECTORY_NAME = "NDSLRecording"
+_GT4PY_SDFG_NAME = "gt4py_sdfg"
+_ORCH_SDFG_NAME = "orch_sdfg"
+
+
+def _upload_to_device(host_data: list) -> None:
```
romanc: duplicated with `optimize.py`

ndsl/dsl/dace/dace_executable.py
```diff
@@ -1,29 +1,208 @@
...
+    AbstractPerformanceCollector,
+    PerformanceCollector,
+)
 from ndsl.quantity import State
 
+DaceExecutables = dict[DaceProgram, "DaceExecutable"]
+DACE_EXECUTABLE_CACHE: DaceExecutables = {}
+
+_BUNDLE_DIRECTORY_NAME = "NDSLRecording"
+_GT4PY_SDFG_NAME = "gt4py_sdfg"
```
romanc: why the reference to gt4py? Imo gt4py would only generate stencils. In my understanding, this is already a full program sdfg (just not optimized). If this is true, could we call it "original" or something?

ndsl/dsl/dace/dace_executable.py
```diff
@@ -1,29 +1,208 @@
...
+
+        self.compiled_sdfg.sdfg.save(
+            f"{bundle_dir}/{_ORCH_SDFG_NAME}.sdfgz", compress=True
+        )
+        with open(bundle_dir / "backend.txt", "w") as f:
+            f.write(self.backend.as_humanly_readable())
+
+    @classmethod
+    def from_serialized_bundle(
+        cls, bundle_dir: str, do_compile: bool = True
```
romanc:
```suggestion
        cls, bundle_dir: str, *, do_compile: bool = True
```

force kwargs just to avoid random `True` / `False`  arguments

ndsl/dsl/dace/dace_executable.py
```diff
@@ -60,5 +239,26 @@ def hash_expected_dsl_args(self, args: tuple[Any], kwargs: dict[str, Any]) -> in
 
         return h
 
+    def replay(self, bench: bool = False) -> None:
```
romanc:
```suggestion
    def replay(self, *, bench: bool = False) -> None:
```

force kwargs just to avoid random `True` / `False`  arguments

ndsl/performance/report.py
```diff
@@ -92,18 +93,24 @@ def gather_timing_data(
...
     return timing_info
 
 
-def write_to_timestamped_json(experiment: Report) -> None:
-    now = datetime.now()
-    filename = now.strftime("%Y-%m-%d-%H-%M-%S")
-    with open(filename + ".json", "w") as outfile:
+def write_to_timestamped_json(experiment: Report, experiment_name: str = "") -> str:
+    now = datetime.now(UTC)
+    filename = experiment_name + now.strftime("%Y-%m-%d-%H-%M-%S") + ".json"
```
romanc: f-string?

tests/dsl/dace/stree/common/test_control_flow.py
romanc, line 1: how about we also move tests to keep the same directory structure under `tests/` as we have under `ndsl/`?

romanc (commented):
This is an awesome refactor. Makes things so much more clear imo. I left a bunch of comments inline; everything from questions about moving functions to simple nitpicks ... 

In general, I think `pathlib.Path` objects could help with a bunch of the current string manipulation of paths, e.g. in `ndsl/dsl/caches/dace.py`. Might be easier to do a follow-up on this, tbh...

ndsl/dsl/caches/__init__.py
FlorianDeconinck: Circular dependency. The entire `cache` system needs rework - I'll move this guy closer to it's only used for now (within `builder`)

ndsl/dsl/caches/dace.py
FlorianDeconinck: Stale comment - strategy is not part of it anymore, `Backend` carries this information.

ndsl/dsl/caches/dace.py
FlorianDeconinck: Fully refactor

ndsl/dsl/caches/dace.py
FlorianDeconinck: Refactored

ndsl/dsl/dace/builder/builder.py
FlorianDeconinck: Moved to `__init__` of `ndsl.dsl.dace`

ndsl/dsl/dace/builder/builder.py
FlorianDeconinck: Refactored

ndsl/dsl/dace/builder/optimize.py
FlorianDeconinck: Refactored

ndsl/dsl/dace/builder/builder.py
FlorianDeconinck: No more `None` return so control flow is keyed upon `is_compiling`

ndsl/dsl/dace/builder/builder.py
FlorianDeconinck: Good catch

ndsl/dsl/dace/builder/parse.py
FlorianDeconinck: Refactored - `parse_sdfg` is now just parsing, all multi-rank & re-entry call logic when parsing is handled in `get_dace_executable`

ndsl/dsl/dace/dace_executable.py
FlorianDeconinck: Renamed

tests/dsl/dace/stree/common/test_control_flow.py
FlorianDeconinck: Test of `dsl.dace` is a mess and exploded in multiple places, apart from the stree ones. Given the time I spent on this I am going to give myself a pass.

Unit tests also needs more work and rework, especially now that parse and optimize and self-standing

FlorianDeconinck:
Deepened the refactor to go all the way to a self-standing `parse_sdfg` and `optimize`. `get_dace_executable` is now the only multi-rank and control flow heavy method.

Added a docstring to `builder` module to spell out things

FlorianDeconinck:
Last remaining bit I'd like to get in is cleaning up the logging

ndsl/dsl/dace/builder/builder.py
```diff
@@ -0,0 +1,101 @@
...
+        if is_compiling:
+            parsed_sdfg = parse_sdfg(
+                dace_program,
+                config,
+                optimization_config,
+                True,  # top most code, since it's the call fetching the executable
+                *args,
+                **kwargs,
+            )
+            original_sdfg = copy.copy(parsed_sdfg)
```
romanc: just checking: it's okay that only the compiling rank gets a reference to the `original_sdfg`, right?

ndsl/dsl/dace/builder/optimize.py
```diff
@@ -0,0 +1,375 @@
+from __future__ import annotations
```
romanc:
```suggestion
```

ndsl/performance/report.py
```diff
@@ -92,18 +93,24 @@ def gather_timing_data(
...
     return timing_info
 
 
-def write_to_timestamped_json(experiment: Report) -> None:
-    now = datetime.now()
-    filename = now.strftime("%Y-%m-%d-%H-%M-%S")
-    with open(filename + ".json", "w") as outfile:
+def write_to_timestamped_json(experiment: Report, experiment_name: str = "") -> str:
+    now = datetime.now(UTC)
+    filename = f"{experiment_name}{now.strftime("%Y-%m-%d-%H-%M-%S")}.json"
```
romanc: locally, I get a `flake8` error on pre-commit because of the quotes ...

```none
ndsl/performance/report.py:110:51: E999 SyntaxError: f-string: unmatched '('
```

switching the outer quotes to single quotes makes it go away 

```suggestion
    filename = f'{experiment_name}{now.strftime("%Y-%m-%d-%H-%M-%S")}.json'
```

not sure what is wrong with CI here ...

ndsl/dsl/dace/builder/parse.py
```diff
@@ -0,0 +1,55 @@
+from __future__ import annotations
```
romanc:
```suggestion
```

ndsl/dsl/dace/orchestration.py
```diff
@@ -1,55 +1,22 @@
...
-from ndsl.dsl.dace.stree.pipeline import StreePipeline
-from ndsl.dsl.dace.utils import (
-    DaCeProgress,
-    memory_static_analysis,
-    report_memory_static_analysis,
-)
-from ndsl.optional_imports import cupy as cp
 from ndsl.quantity import Quantity, State
 
 _INTERNAL__SCHEDULE_TREE_OPTIMIZATION_PASSES: list[tn.ScheduleNodeVisitor] | None = None
```
romanc: left over from moving things around and now unused

```suggestion
```

tests/dsl/dace/stree/common/test_control_flow.py
romanc: sounds good - we can clean up as we go along 👍

romanc (approved):
Awesome. I very much like the new structure with the `cache.py` module inside the builder 💯

ndsl/dsl/dace/builder/builder.py
romanc: well, I guess not, pyFV3 orchestration test is complaining about an unbound variable ...

```none
 UnboundLocalError: cannot access local variable 'original_sdfg' where it is not associated with a value
```

FlorianDeconinck: That was wrong: fixed.

PR #545, opened 2026-08-19
proper deprecation and addition of alternatives

twicki:
# Description

This adds all the new stencil names as we want it while keeping everything else to deprecate.
It duplicates the 2d-stencils in the 2d-stencil file while keeping them in the original. Technically it is breaking to remove them from the 3d file, even though imports should not come form there, but from the top level init and that does support everything still. But we know our code, it happens.

## How has this been tested?

As long as the CI is happy, everything is fine

## Checklist

- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [] My changes generate no new warnings


FlorianDeconinck (approved):
LGTM

PR #549, opened 2026-08-21
[Feature | Opt Cycle II.1] Off-grid tasklet

FlorianDeconinck:
# Description

Off-grid tasklet (tasklet where the inputs/outputs are not indexed with cartesian axis) can block the `AxisMerge` pass (`AxisMerge` knows how to merge only consecutive maps). Prior to this, we used an inline `_push_tasklet_down` within the `AxisMerge`. This has missed cases and added complexity to the code.

We introduce here a series of pass dedicated to dealing with this behavior.
- `ExtractOffGridTasklet`: Gather all off-grid tasklet and store them at the top of the Schedule Tree like a C style declarative process.
- `InlineOffGridTasklet`: Move the off-grid tasklet as-close-as possible to their first usage but_outside_ of any grid tight loop, e.g. C++ optimization style.

Only `ExtractOffGridTasklet` is implemented here.

We also introduce a first pass at SSA with `OffgridTransientScalarSSA` which will allow `ExtractOffGridTasklet` to have shadowing of variable in cases where the scalar written too multiple time.

:warning: This has been coded _defensively_ by limiting the operation to code _outside_ of the cartesian blocks. :warning:

:warning: :warning: This PR deativates `simplify` when doing `stree -> sdfg` due to a bug :warning: :warning:
## How has this been tested?

New utest +
Using the brand new (uncomitted) bencher, we see measurable improvement in D_SW, DivergenceDamping and DelnFlux.

## Checklist


ndsl/dsl/dace/builder/stree/common/control_flow.py
```diff
@@ -1,6 +1,7 @@
 import dace.sdfg.analysis.schedule_tree.treenodes as tn
 
 from ndsl.dsl.dace.builder.stree.common import AxisIterator
+from ndsl.dsl.dace.builder.stree.common.memlet import CARTESIAN_AXIS_SYMBOLS
```
romanc, line 4: Reads weird to import `AxisIterator` and `CARTESIAN_AXIS_SYMBOLS` from two different places. Should we expose `CARTESIAN_AXIS_SYMBOL` in `ndsl.dsl.builder.stree.common` and import both from there?

ndsl/dsl/dace/builder/stree/common/memlet.py
```diff
@@ -136,3 +131,21 @@ def has_dynamic_memlets(first: tn.MapScope, second: tn.MapScope) -> bool:
         ]
     )
     return has_dynamic_memlets
+
+
+class WriteDependencyCollector(tn.ScheduleNodeVisitor):
```
romanc, line 170: Where is this collector used?

ndsl/dsl/dace/builder/stree/common/memlet.py
```diff
@@ -17,11 +16,21 @@ def as_str(self) -> str:
...
-    def is_equal(self, other: str) -> bool:
-        if self == AxisIterator._K:
-            return other.startswith(self.as_str())
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, AxisIterator):
+            return self.value == other.value
+        if isinstance(other, str):
+            if self == AxisIterator._K:
+                return other.startswith(self.as_str())
+            return self.as_str() == other
```
romanc, line 24: With https://github.com/GridTools/gt4py/pull/2752 merged, I think we don't treat the K-axis different anymore:

```suggestion
        if isinstance(other, str):
            return self.as_str() == other
```

Doing so will fail some tests in `tests/dsl/dace/stree/common/test_control_flow.py`

ndsl/dsl/dace/builder/stree/optimizations/off_grid_tasklet.py
```diff
@@ -0,0 +1,83 @@
...
+from ndsl.dsl.dace.builder.stree.common import AxisIterator, list_index
+
+
+class ExtractOffGridTasklet(tn.ScheduleNodeVisitor):
+    """
+    Gather all off-grid tasklet and store them at the top of the Schedule Tree
+    like a C style declarative process.
+
+    Dev note: this pass functions because we keep the tasklet _in order_ the entire
+    time, keeping alive any dependancy on each other correct.
```
romanc:
```suggestion
    time, keeping alive any dependency on each other correct.
```

ndsl/dsl/dace/builder/stree/optimizations/off_grid_tasklet.py
```diff
@@ -0,0 +1,83 @@
...
+            tasklet.parent = node
+
+        # Re-insert in front
+        node.children = [*self._off_grid_tasklets, *node.children]
+
+    def visit_TaskletNode(self, node: tn.TaskletNode) -> None:
+        # Check the inputs are not on-grid
+        for memlet in [*node.in_memlets.values(), *node.out_memlets.values()]:
+            # Are we on-grid
+            if memlet.free_symbols != set():
```
romanc: I don't get this part: shouldn't this somehow check if `memlet.free_symbols` contains references to the grid? The way I read the current implementation is that this check for no data dependency at all, which is correct, but only a subset of what could be labelled as off-grid.

tests/dsl/dace/stree/optimizations/test_off_grid_tasklet.py
```diff
@@ -0,0 +1,110 @@
...
+)
+from ndsl.boilerplate import get_factories_single_tile
+from ndsl.constants import I_DIM, J_DIM, K_DIM
+from ndsl.dsl.gt4py import PARALLEL, computation, interval
+from ndsl.dsl.typing import Float, FloatField
+from tests.dsl.dace.stree import get_SDFG_and_purge
+from tests.dsl.dace.stree.optimizations import Factories
+
+
+def mult_stencils(inout_field: FloatField, scalar: Float):
```
romanc: We have this one as `multiply_to_self(field, factor)`  in `ndsl.stencils`

romanc (changes requested):
Not a huge fan of adding `InlineOffGridTasklet`  when it's not implemented, but I can live with that.

Blocking because I'm not clear on how the on-/off-grid distinction is done. See question inline in `ExtractOffGridTasklet`. Imo this is currently checking if a condition has any data dependency (which imo is just a subset). If we are fine with that subset, I suggest to rename the pass.

ndsl/dsl/dace/builder/stree/optimizations/off_grid_tasklet.py
FlorianDeconinck: It's off _any_ grid now, including `dace` generated `_i0, _i1, _i2` grids. I need to write a pass that collects all the grids on top of the cartesian grids.

But then even with this, it wouldn't be quite correct, because you can pull out tasklet that are on a grid in indexation but not under the `maps` of the grid.

Yeah. I took the more restricted road: no dependency at all on free symbols, which is off grid for sure.

ndsl/dsl/dace/builder/stree/common/memlet.py
FlorianDeconinck: Unused for now but after our discussion I wrote it down because we will need it

ndsl/dsl/dace/builder/stree/optimizations/off_grid_tasklet.py
romanc: I see, okay, we can always tighten this later. So let's keep it as-is.

ndsl/dsl/dace/builder/stree/common/memlet.py
romanc: if we'll use it anyway, I'm fine with merging it unused for now. Just wanted to make sure that we don't accumulate dead code.

romanc (approved):
Thanks for the clarifications and cleanups. Looks all good to me now.

FlorianDeconinck:
Pulling it back - I found some sharp rocks

ndsl/dsl/dace/builder/stree/common/code_block.py
```diff
@@ -0,0 +1,30 @@
...
+from dace.properties import CodeBlock
+from dace.sdfg.analysis.schedule_tree import treenodes as tn
+
+
+def replace_variable_name(code_block: CodeBlock, old_var: str, new_var: str) -> None:
+    """Replace a variable in a CodeBlock using regular expression (full word replacement)"""
+    code_block.as_string = re.sub(rf"\b{old_var}\b", new_var, code_block.as_string)
+
+
+def make_unique_container_name(name: str, root: tn.ScheduleTreeRoot) -> str:
```
romanc: DaCe provides this functionality as `find_new_name` in `dace/utils.py`

https://github.com/spcl/dace/blob/ff05bfc2fc41bb57d1ca1be014eeffd8b4d38cb8/dace/utils.py#L34

ndsl/dsl/dace/builder/stree/optimizations/off_grid_tasklet.py
```diff
@@ -0,0 +1,155 @@
...
+from ndsl import ndsl_log
+from ndsl.dsl.dace.builder.stree.common import AxisIterator, list_index
+from ndsl.dsl.dace.builder.stree.common.code_block import (
+    make_unique_container_name,
+    replace_variable_name,
+)
+from ndsl.dsl.dace.builder.stree.common.memlet import memlet_is_transient_scalar
+
+
+class TransientScalarSSA(tn.ScheduleNodeVisitor):
```
romanc: I think there's a flaw in the logic for renaming reads in subsequent tasklets. I think in your example, the pass would output

```python
    A = tasklet()
    if A:
        f0 = tasklet(A, ...)
    A_0 = tasklet()
    if A_0:                  # correctly renamed
        f1 = tasklet(A, ...) # oops: missing rename A -> A_0
```

The reason I'm saying this is because `visit_Tasklet()` is only looping over the output memlets. Read replacements are only done in `visit_IfScope()`.

The test `reuse_of_scalars` works because the only read is in the `if`, i.e. the subsequent/guarded tasklet is not using `fillc` as an input.

Read before / after write inside the same tasklet is going to be the most fun part, i.e. consider

```py
A = tasklet()
B = tasklet(A)
... # proper usage of B
A = tasklet(A)  # <- this is going to be the most fun tasklet to process
if A:
    C = tasklet(A)
    ...
```

romanc (changes requested):
I think there's a logic flaw in `TransientScalarSSA`, which needs addressing.

ndsl/dsl/dace/builder/stree/common/code_block.py
romanc: Actually, `new_name = sdfg.find_new_symbol(old_name)` is probably what you are looking for. That not only checks arrays, but also symbols and constants.

ndsl/dsl/dace/builder/stree/optimizations/off_grid_tasklet.py
FlorianDeconinck: Yeah I spotted this today debugging something else (sic). So I rename the tasklet inputs too now, it's untested for now.

As for `A = tasklet(A)` - it feels like an order of operations: rename the inputs _before_ you record the name change on a new assign, e.g.

```python
A = tasklet()
B = tasklet(A)
A_0 = tasklet(A)
if A:
    C = tasklet(A_0)
    ...
```

ndsl/dsl/dace/builder/stree/optimizations/off_grid_tasklet.py
```diff
@@ -0,0 +1,206 @@
...
+from ndsl.dsl.dace.builder.stree.common.memlet import memlet_is_transient_scalar
+
+# ----- DEV NOTE -----
+# The pass below have been coded defensively, e.g. they restrict their application to non-cartesian code
+# by skipping the visitor as soon as we detect that we are entering cartesian code
+# ----- DEV NOTE -----
+
+
+class OffgridTransientScalarSSA(tn.ScheduleNodeVisitor):
+    """Transform all off-grid transient scalar throught SSA, e.g.
```
romanc:
```suggestion
    """Transform all off-grid transient scalar through SSA, e.g.
```

ndsl/dsl/dace/builder/stree/optimizations/off_grid_tasklet.py
```diff
@@ -0,0 +1,206 @@
...
+    def __init__(self) -> None:
+        self._ssa_book: dict[str, str] = {}
+
+    def _make_SSA(self, name: str, node: tn.ScheduleTreeNode) -> None:
+        if name not in self._ssa_book:
+            # The first assign - we keep the name
+            self._ssa_book[name] = name
+            return
+
+        candidate = find_new_name(name, list(node.get_root().containers.keys()))
```
romanc: I think, we can just write

```suggestion
        candidate = find_new_name(name, node.get_root().containers)
```

and avoid creating the list of keys: The second argument of `find_new_name` is  used in an `in` (or not) comparison and `node.get_root().containers` is a `dict`. `x in dictionary` will try to find `x` in the keys of `dictionary`, so we should end up with the same answer.

ndsl/dsl/dace/builder/stree/optimizations/off_grid_tasklet.py
```diff
@@ -0,0 +1,206 @@
...
+            self._make_SSA(name, node)
+            # Update the memlet data
+            out_memlet.data = self._ssa_book[name]
+
+    def visit_IfScope(self, node: tn.IfScope) -> None:
+        for in_memlet in node.input_memlets():
+            name = in_memlet.data
+            if name not in self._ssa_book:
+                continue
+            # Update the condtional code and memlet data
```
romanc: typo

```suggestion
            # Update the conditional code and memlet data
```

I think we should do the same in `visit_While` for the condition of the `while` loops.

Not sure if off-grid scalars are ever used in maps / for-loops, but that would be another place that potentially needs replacing. I imagine something like

```py
if flag:
  n_tracers = 3
else:
  n_tracers = 5

for i in range(n_tracers):
  ...
```

but I'm not sure if we have those patterns. maybe in physics codes?

ndsl/dsl/dace/builder/stree/optimizations/off_grid_tasklet.py
```diff
@@ -0,0 +1,206 @@
...
+        for child in node.children:
+            self.visit(child)
+
+    def visit_TaskletNode(self, node: tn.TaskletNode) -> None:
+        # Check the inputs are not on-grid
+        for memlet in [*node.in_memlets.values(), *node.out_memlets.values()]:
+
+            # Are we on-grid, e.g. are we dependent on grid indexation
+            #
+            # Dev NOTE: because we can have _many_ grids due to DaCe parsing some pythong
```
romanc:
```suggestion
            # Dev NOTE: because we can have _many_ grids due to DaCe parsing some python
```

ndsl/dsl/dace/builder/stree/optimizations/off_grid_tasklet.py
```diff
@@ -0,0 +1,206 @@
...
+
+        self._off_grid_tasklets.append(node)
+
+
+class InlineOffGridTasklet(tn.ScheduleNodeVisitor):
+    """
+    Move the off-grid tasklet as-close-as possible to their first usage but
+    _outside_ of any grid tight loop, e.g. C++ optimization style.
+
+    Dev note: movement need to be made in reverse order so we can keep any dependancy on
```
romanc:
```suggestion
    Dev note: movement need to be made in reverse order so we can keep any dependency on
```

ndsl/dsl/dace/builder/stree/optimizations/off_grid_tasklet.py
```diff
@@ -0,0 +1,206 @@
...
+
+            # Are we on-grid, e.g. are we dependent on grid indexation
+            #
+            # Dev NOTE: because we can have _many_ grids due to DaCe parsing some pythong
+            # into its own many-maps we go more stricter here and we restrict to
+            # any symbols that are not resolved yet (and therefore could be indexer).
+            # The fix is to create a pass that collects all the grids, and change this check
+            # to look into the "indexer symbols" as well as the topology to see if we are indeed
+            # under a grid calculation (e.g. under or after maps indexed)
+            if memlet.free_symbols != set():
```
romanc: I find the following easier to parse as "if has any free symbols"

```suggestion
            if len(memlet.free_symbols) > 0:
```

but works either way. Feel free to discard.

tests/dsl/dace/stree/optimizations/test_merge.py
```diff
@@ -47,6 +47,16 @@ def stencil_with_buffer_read_offset_in_K(
...
 
 
+def stencil_with_buffer_read_offset_in_Kp1(
+    in_field: FloatField, out_field: FloatField, buffer: FloatField
+) -> None:
+    with computation(PARALLEL), interval(1, None):
+        buffer = in_field + 6
+
+    with computation(PARALLEL), interval(1, None):
+        out_field = buffer[K + 1] + 7
```
romanc: Looks like a straight copy from the `K - 1` case. With `K + 1` here, shouldn't the bounds of the interval also change? I'd expect to see `interval(0, -1)` to avoid reading out of bounds for `K+1` at the last level.

romanc: Actually, this stencil isn't used, right? Is that on purpose?

tests/dsl/dace/stree/optimizations/test_off_grid_tasklet.py
```diff
@@ -0,0 +1,183 @@
...
+)
+from ndsl.boilerplate import get_factories_single_tile
+from ndsl.constants import I_DIM, J_DIM, K_DIM
+from ndsl.dsl.gt4py import PARALLEL, computation, interval
+from ndsl.dsl.typing import Float, FloatField
+from tests.dsl.dace.stree import get_SDFG_and_purge
+from tests.dsl.dace.stree.optimizations import Factories
+
+
+def mult_stencil(inout_field: FloatField, scalar: Float):
```
romanc:
```suggestion
def mult_stencil(inout_field: FloatField, scalar: Float) -> None:
```

tests/dsl/dace/stree/optimizations/test_off_grid_tasklet.py
```diff
@@ -0,0 +1,183 @@
...
+        self, scalar: float, in_field: FloatField, out_field: FloatField
+    ) -> None:
+        for n in unroll(range(2)):
+            fillc = self._fillc_value[n]
+            if fillc:
+                self._mult_stencil(in_field, scalar)
+
+    def reuse_of_scalars_in_inputs(
+        self, scalar: float, in_field: FloatField, out_field: FloatField
+    ):
```
romanc:
```suggestion
    ) -> None:
```

tests/dsl/dace/stree/optimizations/test_off_grid_tasklet.py
```diff
@@ -0,0 +1,183 @@
...
+        tmp_scalar = scalar * 2.0
+        self._mult_stencil(in_field, tmp_scalar)
+        tmp_scalar = scalar * 2.0
+        self._mult_stencil(in_field, tmp_scalar)
+
+    def block_by_conditional(
+        self, scalar: float, in_field: FloatField, out_field: FloatField
+    ):
```
romanc:
```suggestion
    ) -> None:
```

romanc (approved):
Very happy to see this coming 🚀 

I found a corner case inline with the SSA pass. Feel free to defer the work to a follow-up if you think it's not likely to make problems now.

ndsl/dsl/dace/builder/stree/optimizations/off_grid_tasklet.py
FlorianDeconinck: We have this pattern in the stencils, which is one reason I excluded all the cartesian computation for it. I can definitely see that happening, it's a  very common runtime config switch.

I am not sure how to resolve it though. I guess I have to read the actual literature on SSA instead of winging it.

romanc: I see ... let's do it in a follow-up then.

tests/dsl/dace/stree/optimizations/test_merge.py
FlorianDeconinck: I wrote a bunch and then made the tests but ended up not using it. Considering how the offset analysis is done now K + 1 or K - 1 is equivalent. When we do the serial + merging of `for` properly, this might become handy. Removing for now.

ndsl/dsl/dace/builder/stree/optimizations/off_grid_tasklet.py
FlorianDeconinck: I can't really leave this is as is, it's going to bite us hard... I am going to try and fix it. It could just be "only rename if there's a read in between" ?

FlorianDeconinck: We have restricted merge to only apply to tasklet with scalar transient outputs that are SSA properly.

The above case requires a `Phi` function and the current pass doing the SSA does not apply `Phi` functions. Therefore the extraction will be skipped and we are safe.

ndsl/dsl/dace/builder/stree/optimizations/off_grid_tasklet.py
```diff
@@ -79,36 +96,83 @@ def visit_TaskletNode(self, node: tn.TaskletNode) -> None:
...
 
 class ExtractOffGridTasklet(tn.ScheduleNodeVisitor):
     """
-    Gather all off-grid tasklet and store them at the top of the Schedule Tree
+    Gather all off-grid nodes and store them at the top of the Schedule Tree
     like a C style declarative process.
 
-    Dev note: this pass functions because we keep the tasklet _in order_ the entire
+    Dev note: this pass functions because we keep the node _in order_ the entire
     time, keeping alive any dependency on each other correct.
```
romanc:
```suggestion
    Dev note: this pass functions because we keep the nodes _in order_ the entire
    time, keeping any dependency on each other correct.
```

ndsl/dsl/dace/builder/stree/optimizations/off_grid_tasklet.py
```diff
@@ -177,15 +282,21 @@ def visit_TaskletNode(self, node: tn.TaskletNode) -> None:
             if memlet.data in self._on_grid_data:
                 return
 
-        self._off_grid_tasklets.append(node)
+            # We check that the output is SSA, e.g. written a single time throughout
+            # if not then we will
```
romanc: if not, then we will not extract the tasklet?

romanc (approved):
I think this is looking good 🥳

ndsl/dsl/dace/builder/stree/optimizations/off_grid_tasklet.py
FlorianDeconinck: I like my code to have some suspense. We will do something, but what? You'll need to wait for the next PR to know!

romanc: In that case, I can only hope there's gonna be a sequel and this isn't one if those cliff hangers that never get resolved because the next season was cancelled ...

PR #554, opened 2026-08-27
build: add support for python 3.14

romanc:
# Description

This PR brings python 3.14 support for NDSL. In general, this looks fine: CI shows one new warning (deprecated function call in a dependency - not for us to do anything about). [pyFV3 CI](https://github.com/NOAA-GFDL/pyFV3/pull/161) also seems to run fine with python 3.14 (same new deprecation note).

In manual testing, I found two gotchas

1. `zarr` only has official support for 3.14 starting with 3.2.0. We currently limit it to < 3 because of issue. This gives us old dependencies of `zarr`. In particular, `numcodecs` 0.15.1 will not work with gcc >= 15 because of [a bad typedef](https://github.com/Blosc/c-blosc/issues/393). Works okay on GitHub Actions because our runners [only have gcc 12, 13, and 14](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md).
2. `pySHiELD` depends on a "NOAA-version" of `pyrte-rrtmgp`, which is not up-to-date with the upstream `pyrte-rrtmgp`. The "NOAA-version" only supports python version 3.11 and 3.12. The upstream version also supports 3.13 and 3.14. This means both pySHiELD and pace won't run in 3.13 (and 3.14) until someone pull the upstream changes.

Issue 1 is a bit annoying and might be blocking depending on how urgent we see this. Issue 2 is just for NOAA to update their home-grown version of `pyrte-rrtmgp`.

## How has this been tested?

- CI of NDSL (see below) and [pyFV3](https://github.com/NOAA-GFDL/pyFV3/pull/161)
- Manual install on my machine

## Checklist

  New deprecation warning in dependency. Nothing for us to do except eventually updating.


FlorianDeconinck:
`zarr` is used for a single purpose: a slightly more compressed `Monitor`. 

@oelbert do you use the `ZarrMonitor` ?

romanc:
related issue: https://github.com/NOAA-GFDL/NDSL/issues/157

romanc:
Looks like this is used in `pace` diagnostics, but only as an option next to the netCDF option

https://github.com/NOAA-GFDL/pace/blob/5853e8a7e8286e5b9cca1b3a6099304005efedd5/pace/diagnostics.py#L118-L124

@oelbert would it be okay to remove `zarr`  support from `pace`?

romanc:
Looks like the `ZarrMonitor` is pretty deep in `pace` e.g. used as default for many examples.

[Zarr v3 update guide](https://zarr.readthedocs.io/en/stable/user-guide/v3_migration/) for reference. Looks like a significant refactor. To be evaluated (on Monday) if we rely on any of the "Unimplemented Zarr-Python 2 features"...

romanc:
Okay, so I went and just did the update to `zarr` version 3. Was kind of okay. I hope, I got everything. NDSL tests were green again locally.

The additional `numcodes` dependency is just to avoid problems with GCC >=15. I have an open PR in `zarr` (see https://github.com/zarr-developers/zarr-python/pull/4299) to fix that properly. Let's hope we get that into `zarr` version 3.3.1 (which is apparently due soon).

`pace` tests are still red because `zarr` has breaking changes that need to be coordinated with the NDSL update: NDSL won't be able to do the `zarr` update without breaking changes because `zarr` has - well - breaking changes. So I'm afraid, this needs to be a coordinated effort where the `zarr` update in `pace` needs to happen at the same time as the update from NDSL 2026.08.00 to NDSL 2026.09.00.

